PR #3, opened 2015-07-15
Support Pandoc's fancy_lists extension

zklhp:
Hi Jason,
The Pandoc version of Markdown has a fancy_lists extension (see http://pandoc.org/README.html#ordered-lists), which means `#` can be used as ordered list marker. At present the markdown-mode fontified `#. blah blah` as header. This is inconvenient. This pull request makes markdown-mode recognize `#.` not as header marker but as list marker. I hope this can be accepted. Thanks.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1195,7 +1195,7 @@ Group 5 matches the reference label.")
   "Regexp identifying Markdown headers.")
 
 (defconst markdown-regex-header-1-atx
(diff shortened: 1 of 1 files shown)


jrblevin:
Thanks for this patch. It broke one of the test cases regarding ATX headers, so I had to use a slightly different regular expression. I also added support for filling "fancy lists" and test cases for font lock and filling. See commit 4b8ac273.

PR #8, opened 2015-07-30
handle ikiwiki-style search for wikilinks

csrhodes:
ikiwiki (at <http://ikiwiki.info/>) supports searching for wikilink
targets not just relative to the current directory, but also relative to
parent directories (up to the wiki root).

This patch supports that search unconditionally.  It also goes beyond
the wiki root, because the emacs mode has no way of knowing where that
root is.  I hope that I have got the termination logic right but the
platform where it's most likely to be wrong (Windows) is currently
inaccessible to me.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -4584,10 +4584,20 @@ and [[test test]] both map to Test-test.ext."
     (when (eq major-mode 'gfm-mode)
       (setq basename (concat (upcase (substring basename 0 1))
                              (downcase (substring basename 1 nil)))))
-    (concat basename
-            (if (buffer-file-name)
-                (concat "."
-                        (file-name-extension (buffer-file-name)))))))
+    (let* ((default
+            (concat basename
+                    (if (buffer-file-name)
+                        (concat "."
+                                (file-name-extension (buffer-file-name))))))
+           (current default))
+      (catch 'done
+        (loop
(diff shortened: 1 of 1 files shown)


csrhodes:
I should perhaps say that this isn't perfect by any means:

* it should also check in a child directory named the same as the current markdown file (without extension)
* it needs to be a bit more robust in abandoning (so that e.g. complicated ikiwiki directrives beginning e.g. `[[!template ...]]`don't get treated as filenames in some emacs-native syntax such as TRAMP.

jrblevin:
Thanks for this patch. I [used to][1] [use][2] [Ikiwiki][3] myself, so I remember how this works. I've applied your patch (f81bfdd), but I think it might be best to make it optional since this kind of search behavior is somewhat unique. In a follow-up commit (4a598b2), I added some documentation as well as a custom variable to disable it.

Finally, a test case or two would be very helpful to ensure that the new behavior doesn't get inadvertently broken in the future. For example, a basic test would create two nested directories, each containing a Markdown file, with the deeper file having a wiki link back to the higher level file, then check to see that finding the file works.

[1]: http://jblevins.org/projects/ikiwiki/
[2]: https://ikiwiki.info/users/jasonblevins/
[3]: https://ikiwiki.info/

PR #15, opened 2015-09-01
Console Emacs -friendly backspace/tab bindings

kini:
This commit fixes a couple of bindings so that they work in console
Emacs (i.e. `emacs -nw`) as well as GUI Emacs.

Incidentally, this also improves interoperability with evil-mode,
since before this commit <backspace> remained bound in evil's Normal
State (where usually <backspace>, or rather DEL, would be bound to
`evil-backward-char'.


jrblevin:
Thanks!

PR #21, opened 2015-09-10
give wiki-link variables safety annotations

csrhodes:
these wiki-link variables are potentially customized per-file or
per-project (in e.g. a .dir-locals.el file).  Specify that they are
`safe' to have boolean values.


jrblevin:
Applied--thanks for the patch!

PR #23, opened 2015-09-10
Ikiwiki tests

csrhodes:
an attempt at writing tests for the ikiwiki-like parent directory search for wikilink targets.

I got thoroughly lost by font-lock weirdness when writing this; differences in behaviour when run interactively and in batch mode are weird.  (Even without my modifications, I think a number of tests
fail when run in a live emacs with `ert-run-tests-interactively`, so...)

Nevertheless, with this `make test` succeeds.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -4739,7 +4739,8 @@ and [[test test]] both map to Test-test.ext."
                              (downcase (substring basename 1 nil)))))
     (let* ((default
             (concat basename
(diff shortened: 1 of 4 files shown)


jrblevin:
Thanks for these patches. Sorry it took me so long to apply them. I have the same issue when running the tests interactively, but with `make test` your new test passes for me as well. Committed as 9848e3f and 5aa7380.

PR #32, opened 2015-10-11
add asymmetric header styling option, which most folk use these days

jonmountjoy:
This creates a new customization variable, `markdown-asymmetric-header`.  Set to non-nil to only have header characters to the left of the title. The default will ensure header characters are placed to the left and right of the title.

For example, if set, headers will be styled like so:

```
### foo bar
```

instead of like this

```
### foo bar ###
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,6 +441,10 @@ provides an interface to all of the possible customizations:
   * `markdown-italic-underscore` - set to a non-nil value to use
     underscores for italic instead of asterisks (default: `nil`).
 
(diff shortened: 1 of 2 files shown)


blaenk:
Great!

vyp:
Thank you @jonmountjoy.

jrblevin:
Thanks, indeed! Applied just now...

jonmountjoy:
Cheers!

PR #34, opened 2015-10-28
Make (beginning-of-defun -1) go to next title when point is at beginning of defun

Ambrevar:


jrblevin:
Thanks for catching this. Applied as 92fc302 with new tests in 4f75107.

Issue #33, opened 2015-10-26
Lines beginning with hash symbol in literal code blocks treated as headers


duncanburke:
For the purposes of commands such as `outline-next-visible-heading` and `outline-forward-same-level`, it is not expected that lines beginning with the `#` symbol in literal code blocks be treated as headers.

For example:

    ### Third level ###

    * A list item
    * and another
      with some overhang
       + and a sublist
         with overhang
    - And a third with a different marker

    ### Third level number two ###

    ~~~ {.bash}
    #!/bin/bash

    # In a code block, this shouldn't be a header
    ~~~

    ### Level two again

    followed by some body text

jrblevin:
I think the new `markdown-collect-fm-code-blocks` function introduced in #40 could be used for font lock to fix this issue also. By identifying fenced code blocks first, then identifying headings that are not already marked as code blocks. I'll try to take a look soon, but if anyone can get to it sooner please submit a pull request.

syohex:
In `outline-next-visible-heading`, outline-mode uses only regular expression(`outline-regexp`) for searching header. We can control its behaviour only by `outline-regexp`. So we need to write wrapper command for avoiding this issue as below.

```lisp
(defun markdown-move-visible-header-common (move-fn arg)
  (let ((blocks (markdown-collect-gfm-code-blocks)))
    (funcall move-fn arg)
    (while (markdown-gfm-code-block-p blocks (point))
      (funcall move-fn arg))))

(defun markdown-next-visible-heading (arg)
  (interactive "p")
  (markdown-move-visible-header-common #'outline-next-visible-heading arg))

(defun markdown-next-visible-heading (arg)
  (interactive "p")
  (markdown-move-visible-header-common #'outline-previous-visible-heading arg))
```

I don't understand `outline-forward-same-level` behavior yet :-(

syohex:
I sent PR #41. Please check it.

jrblevin:
Thanks for this report, @duncanburke, and thanks @syohex for the patch. As I mentioned in #41 I have been working on syntax table support for code blocks. In addition to helping with font lock, this is also related to the heading movement commands. I used #41 as a basis for the changes. So, this should be fixed in 5daa9fc.

PR #35, opened 2015-10-29
Modified markdown-regex-header to require a space between # and heading text

plantarum:
When a markdown document contains a code block for a language that uses # as a comment character, folding is broken. For example, in the following markdown text:

```
# Heading One

` ``r
#Comment on code
myFunction <- function
` ``

## Subheading A
```
Folding the first header produces the following:

```
# Heading One
#Comment on code
myFunction <- function
` ``

## Subheading A
```

In effect, my code comment is treated as a heading for the folding algorithm.

A simple fix is updating the heading regexp to **require** a space (\s-+), rather than allow a space (\s-*), after the opening # symbols. This appears to be consistent with the markdown specification, in which all headlines are of the form "# Heading 1" rather than "#Heading 1". 

As a consequence, in code blocks we cannot use a space between the # and the code. The alternative would require detecting and ignoring lines that occur in a code block, which would be more involved.

Thanks for your time,

Tyler

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1302,7 +1302,7 @@ Group 2 matches only the label, without the surrounding markup.
 Group 3 matches the closing square bracket.")
 
 (defconst markdown-regex-header
-  "^\\(?:\\(.+\\)\n\\(=+\\)\\|\\(.+\\)\n\\(-+\\)\\|\\(#+\\)\\s-*\\(.*?\\)\\s-*?\\(#*\\)\\)$"
+  "^\\(?:\\(.+\\)\n\\(=+\\)\\|\\(.+\\)\n\\(-+\\)\\|\\(#+\\)\\s-+\\(.*?\\)\\s-*?\\(#*\\)\\)$"
   "Regexp identifying Markdown headers.")
 
 (defconst markdown-regex-header-1-atx


vyp:
> As a consequence, in code blocks we cannot use a space between the # and the code.

So we cannot have a comment character `#`, then a space, and then the comment? i.e. `# Comment on code`? I don't think that would be a good solution.

Also, maybe off topic, but why do you have a space in your three backticks? Is that valid, or is that a typo?

plantarum:
I agree this is not ideal. But to be clear, right now, with no changes, comments with or without a space in source code blocks are treated as markdown headings. My change would mean that comments without a space would be ignored. I'm not introducing a new problem, I'm partially fixing an existing problem.

It would be much better to solve the problem entirely, but as I mentioned above, that requires more work in the outline-mode code responsible for section folding. I'm not familiar enough with that to know how easy or hard it would be.

I put a space in my backticks so they would get displayed as backticks, and not interpreted as the end of the code block. It's not correct syntax, but close enough to convey the issue I think. I tried escaping the backticks with a backslash, but that didn't work.

vyp:
> I'm not introducing a new problem, I'm partially fixing an existing problem.

Yes I see now this is correct, thanks for clarifying. (I guess once/if #33 is fixed, this change can be reverted.)

> I put a space in my backticks so they would get displayed as backticks, and not interpreted as the end of the code block.

Ah I see. You could always indent your block by four spaces to do this (instead of using backticks):

    # Heading One

    ```r
    #Comment on code
    myFunction <- function
    ```

    ## Subheading A

This will not do syntax highlighting though. But at least you can convey the text correctly.

plantarum:
Yes, I hadn't seen issue 33, but that is exactly the problem my PR addresses, in a clumsy and incomplete way.

jrblevin:
Indeed, this is related to #33. The folding and imenu code needs to be rewritten to ignore false positives (lines appearing to be ATX headers) inside fenced code blocks.

I agree that having a space between the hash mark and heading text is stylistically preferred. However, I allowed the space to be optional so that markdown-mode would be as permissive as possible and maintain compatibility with Markdown.pl and the [vast majority](http://johnmacfarlane.net/babelmark2/?normalize=1&text=%23A) of Markdown processors that do allow headings without spaces.

PR #36, opened 2015-11-04
add markdown-native-preview-mode to view previewed output within emacs

cosmicexplorer:
`markdown-export-and-preview` is great for running glossy live-updating markdown viewer nightmare children and is very idiomatic emacs lisp. `markdown-native-preview-mode` is an alternative using `eww` (built in to emacs since 24.4) to view compiled HTML output within emacs, live-updating as the file is saved. The function `markdown-export-native-preview` will run `markdown-export`, then open `eww` to the exported file. If `eww` is not already displaying a preview of the exported output, it will pop to a new `eww` buffer displaying it; otherwise, it will update all windows displaying the exported output to their respective points in a new eww buffer rendering the new export. If `markdown-native-preview-mode` is turned on, rendering occurs synchronously on every save.

The defcustom `markdown-preview-window-function` can be used to customize the method of displaying the exported output in emacs, allowing the user to choose an alternative html viewer within emacs such as `w3m`. The defcustom `markdown-native-preview-delete-export` determines whether to delete the exported file, which is typically not required after rendering within emacs, unlike with external html viewers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,10 +208,14 @@ keybindings by pressing <kbd>C-c C-h</kbd>.
     and save the result in the file `basename.html`, where
     `basename` is the name of the Markdown file with the extension
     removed.  *Export and View:* press <kbd>C-c C-c v</kbd> to export the
-    file and view it in a browser.  **For both export commands, the
-    output file will be overwritten without notice.**
-    *Open:* <kbd>C-c C-c o</kbd> will open the Markdown source file directly
-    using `markdown-open-command`.
+    file and view it in a browser.  *Open:* <kbd>C-c C-c o</kbd> will open
+    the Markdown source file directly using `markdown-open-command`.
+    *Live Export*: Press <kbd>C-c C-c l</kbd> to turn on
+    `markdown-live-preview-mode` to view the exported output
+    side-by-side with the source Markdown. **For all export commands,
+    the output file will be overwritten without notice.**
+    `markdown-live-preview-window-function` can be customized to open
+    in a browser other than `eww`.
 
     To summarize:
 
@@ -221,6 +225,7 @@ keybindings by pressing <kbd>C-c C-h</kbd>.
       - <kbd>C-c C-c v</kbd>: `markdown-command` > `basename.html` > browser.
       - <kbd>C-c C-c w</kbd>: `markdown-command` > kill ring.
       - <kbd>C-c C-c o</kbd>: `markdown-open-command`.
+      - <kbd>C-c C-c l</kbd>: `markdown-live-preview-mode` > `*eww*` buffer.
 
     <kbd>C-c C-c c</kbd> will check for undefined references.  If there are
     any, a small buffer will open with a list of undefined
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1008,12 +1008,9 @@ to `nil' instead. See `font-lock-support-mode' for more details."
   :group 'markdown
   :type 'boolean)
 
-(defcustom markdown-live-preview-window-function 'eww-open-file
-  "Function to display preview of markdown output within emacs. Function must
-emulate the behavior of `eww-open-file' in that it must create a new buffer
-containing the preview of the html file, and then call `switch-to-buffer' to
-that buffer so that upon completion, `selected-window' displays the buffer
-previewing the exported html."
+(defcustom markdown-live-preview-window-function 'markdown-live-preview-window-eww
+  "Function to display preview of Markdown output within Emacs. Function must
+update the buffer containing the preview and return the buffer name."
   :group 'markdown
   :type 'function)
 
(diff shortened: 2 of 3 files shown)


jrblevin:
Thanks, I really like this idea. I didn't know about eww previously. I didn't have a whole lot of time to test it this morning but I did notice a few issues:

1. These changes break the `test-markdown-hook/after-export` test. I didn't inspect further to see why.

2. When editing a buffer with `markdown-native-preview-mode` on, when I save the buffer the `*eww*` preview does indeed update but then the buffer I'm editing gets killed (i.e., it is closed and I'm back to `*scratch*`). I don't think that's the intention.

3. `markdown-export-native-preview` and `markdown-native-preview-mode` don't open the `*eww*` buffer (not even a half window). Should they? I'm not sure, just raising it as a possibility to help with "discoverability".

4. A short description added to the documentation at the top of the file (which is used to produce README.md) would also be helpful. Eventually, a keybinding would be nice.

If you can take a look at 1 and 2 (3 and 4 can wait), then I'd be very happy to merge this.

cosmicexplorer:
For some reason, I don't see the `after-export` test breaking, but I do see a variety of others failing. I'll investigate this. I believe 2. is a result of some sketchy buffer/window-handling in the export function, which I'll fix. For 3., it should should be calling display-buffer on the `*eww*` buffer, so I don't know why that's happening (I think it's because of some sketchy things I'm doing in my personal emacs config with `eww` buffers). 4. is easy, and I'll do that.

jrblevin:
Somehow the font lock tests aren't reliable when running interactively. I run `make test` on the command line instead. One of the conditions checked in the `after-export` test is that the output buffer remains open, but it is now failing because of the following line added to `markdown-export`:

    (kill-buffer output-buffer)

Also, just to make sure it wasn't due to my local config, I double-checked 2 and 3 above with `emacs --no-init-file` and nothing changed.

cosmicexplorer:
Gotcha, the tests work now (some obsoleted functions in 25.1 tripped me up but that's just cause I'm running from master). I just added that because I personally don't like having the exported html open in a buffer after I'm done, but that's the job of `markdown-export-native-preview`, not `markdown-export`, so I've moved it to my function.

I've also made a few changes to the use of `font-lock-fontify-buffer`; my version of emacs was yelling about not being allowed to use them non-interactively when running `make-test`. This shouldn't break anything on your end.

Please tell me if the markdown buffer is still being killed or if the changes I made to `test/markdown-test.el` break anything.

cosmicexplorer:
Hey, so I've checked this out on `emacs -Q` and my own configuration; 3 should be fixed now (it should use `display-buffer` to pop up the `eww` buffer). Since you said you tested it on `--no-init-file`, I'm confused as to why it's not popping up in your testing. Can you check if the new version fails?

cosmicexplorer:
Renamed `native-preview` -> `live-preview`, because "native preview" doesn't actually mean anything and <kbd>C-c C-c n</kbd> was already taken. Added a bit of documentation, keybinding, and menu item.

jrblevin:
This is looking great. I'm still seeing some window issues on Emacs 24, however. Now, when I enable `live-preview` with <kbd>C-c C-c l</kbd> the `*eww*` buffer flashes very briefly but disappears. A second issue is that when I disable `live-preview`, by pressing <kbd>C-c C-c l</kbd> again, the buffer for the Markdown file I am editing gets closed.

I tested this on a recent build of Emacs 25 and it works as it should. The `*eww*` buffer is opened as a half window and gets closed when I turn off `live-preview`. I'm not up to speed on the new developments in Emacs 25, so I don't have any idea why the behavior is different.

jrblevin:
For some reason `markdown-preview-buffer` isn't getting set correctly on Emacs 24. It references the buffer for the Markdown file rather than the `*eww*` buffer. Perhaps it's a difference in `eww-open-file`?

jrblevin:
I simplified the window handling code and refactored the live-preview-window-function definition so that it works for me now on Emacs 24 and 25. See commit 97cae6e in my `feat/live-preview` branch.

I admit that I didn't fully understand what the additional window handling code was doing. I'm simply using `switch-to-buffer-other-window` on the first call, and doing nothing on subsequent updates (leaving it to the user to keep the window around). What do you think?

cosmicexplorer:
Took me a bit, but I figured out that the complex window handling code was because in my own config I change the name of the `*eww*` buffer after it's created, which was causing part of the confusing window management code. I'm definitely ok with removing that, but I'd prefer it if `markdown-window-redisplay-preview` was kept, because it keeps the window scrolled to the same spot every time the screen is refreshed. This should keep the spirit of your changes, but also ensure that we don't modify the point/start of any windows displaying the output buffer (for me, whenever I refresh the `*eww*` buffer, it scrolls all the way to the bottom on every save unless I do this).

I also added the export function to `after-revert-hook`, and killed the attached preview buffer on `kill-buffer-hook`. Those can be removed or `defcustom`ed as required.

cosmicexplorer:
Removed the `after-revert-hook` because I forgot that it turns off minor modes anyway. Checked this on a clean emacs 24 and 25, and both seemed to work correctly. Thanks for putting so much time into this.

PR #41, opened 2015-12-09
Improve moving header commands in GFM

### LINKED ISSUE #33: Lines beginning with hash symbol in literal code blocks treated as headers

syohex:
Ignore header like statements in code blocks.

I suppose this patch fixes #33 issue. Please review.

markdown-mode

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -3554,8 +3554,8 @@ Assumes match data is available for `markdown-regex-italic'."
     (define-key map (kbd "<S-tab>")  'markdown-shifttab)
     (define-key map (kbd "<backtab>") 'markdown-shifttab)
     ;; Header navigation
-    (define-key map (kbd "C-c C-n") 'outline-next-visible-heading)
-    (define-key map (kbd "C-c C-p") 'outline-previous-visible-heading)
+    (define-key map (kbd "C-c C-n") 'markdown-next-visible-heading)
+    (define-key map (kbd "C-c C-p") 'markdown-previous-visible-heading)
     (define-key map (kbd "C-c C-f") 'outline-forward-same-level)
     (define-key map (kbd "C-c C-b") 'outline-backward-same-level)
     (define-key map (kbd "C-c C-u") 'outline-up-heading)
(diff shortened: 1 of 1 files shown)


jrblevin:
Many thanks for this patch. It works for moving between the previous/next headers (of any level), but not yet for moving between headers of the same level (for example, if at level 2, it skip any 3+ headers and move to the next level 2). This gave me some ideas for rewriting the other outline-mode function, so haven't committed it yet, but I hope to work it in as part of a broader patch soon.

jrblevin:
I'm also working on test cases, etc., and I had one possible concern: I wonder if there might be some unexpected consequences of setting code blocks to be level 7 in the `markdown-outline-level` function. I understand the basic logic, since there are at most six levels of ATX-style headings, but this function is also called by some of the font lock and block movement commands. I want to test it a bit more.

syohex:
> but not yet for moving between headers of the same level (for example, if at level 2, it skip any 3+ headers and move to the next level)

Please tell me what's wrong in more detail. I suppose this patched version commands works like such behaviour.

jrblevin:
In other words, the patch works for `markdown-next-visible-heading` and `markdown-previous-visible-heading` but not for `outline-forward-same-level` and `outline-backward-same-level`. Thank you for the patch, but rather the commit a partial fix I am working on a larger rewrite of the code block parsing functions that should improve font lock, heading folding, heading movement, etc. with respect to headers inside code blocks. My plan is to mark code blocks at the syntax analysis stage and then in the font lock and movement stages, we can simply look at the syntax properties rather than rescanning the document for all code blocks. Stay tuned...

PR #43, opened 2015-12-18
Add Travis CI configuration

syohex:
Don't test Emacs 24.3 or lower version. Because eww was introduced at
Emacs 24.4 and byte-compile test is failed on older version.

Test result sample is here
- https://travis-ci.org/syohex/markdown-mode

You need to setup for enabling CI of this repository from your travis profile page.

See also
- https://docs.travis-ci.com/user/getting-started/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: generic
+sudo: false
+before_install:
+  - curl -fsSkL https://gist.github.com/rejeep/ebcd57c3af83b049833b/raw > x.sh && source ./x.sh
+  - evm install $EVM_EMACS --use --skip
+env:
+  - EVM_EMACS=emacs-24.3-travis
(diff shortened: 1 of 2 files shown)


jrblevin:
This is useful, thanks! I will set this up after fixing some outstanding bugs. Also good to know that we need to require Emacs 24.4 for eww. It would be good to work on backwards compatibility (i.e., disabling live preview mode) for previous versions.

syohex:
I updated for passing tests on Emacs 24.3. Result is here(https://travis-ci.org/syohex/markdown-mode). markdown-mode uses `defvar-local` which was introduced at Emacs 24.3. So Emacs 24.1 and 24.2 cannot byte-compile markdown-mode.el.

jrblevin:
I applied these commits and also added a badge to README.md. Many thanks!

https://travis-ci.org/jrblevin/markdown-mode

PR #45, opened 2015-12-19
add more nuance to markdown-live-preview-delete-export

cosmicexplorer:
features like cross-references (#refs) don't work in
markdown-live-preview-mode, because it deletes the output html file on
each export. this can be turned off, but then an html file is left over
after an editing session, and has to be manually deleted. this adds the
ability to only delete the output file automatically after quitting a
markdown-live-preview-mode session.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -5096,8 +5096,8 @@ the rendered output."
   (when (buffer-live-p markdown-live-preview-buffer)
     (kill-buffer markdown-live-preview-buffer))
   (setq markdown-live-preview-buffer nil)
-  (when (or (eq markdown-live-preview-delete-export 'delete-on-export)
-            (eq markdown-live-preview-delete-export 'delete-on-destroy))
+  (when (memq markdown-live-preview-delete-export
+              '(delete-on-export delete-on-destroy))
     (let ((outfile-name (markdown-live-preview-get-filename)))
       (when (file-exists-p outfile-name)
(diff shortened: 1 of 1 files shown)


markdown-mode.el
```diff
@@ -5089,7 +5095,12 @@ the rendered output."
 (defun markdown-live-preview-remove ()
   (when (buffer-live-p markdown-live-preview-buffer)
     (kill-buffer markdown-live-preview-buffer))
-  (setq markdown-live-preview-buffer nil))
+  (setq markdown-live-preview-buffer nil)
+  (when (or (eq markdown-live-preview-delete-export 'delete-on-export)
+            (eq markdown-live-preview-delete-export 'delete-on-destroy))
```
syohex: This part can be simplified by `memq` as below.

```
(memq markdown-live-preview-delete-export  '(delete-on-export delete-on-destroy))
```

jrblevin:
Thanks for confirming. It's minor, I know, but I wanted to make sure I understood the new functionality. I will merge this in. Thanks again for the patches!

jrblevin:
Done. I took the liberty of squashing the four commits and rebasing on the latest HEAD. Hope you don't mind. Thanks again!

By the way, I'm planning a new stable release soon. There has been a lot of activity lately, so let me know if you notice any issues.

cosmicexplorer:
Awesome! I'll check out the new stuff, see if it all works out together.

PR #46, opened 2015-12-24
delete html editing buffer for markdown-live-preview-export

cosmicexplorer:
The buffer containing the exported version of the markdown file (`<file>.html`) is not killed when exported using `markdown-live-preview-export`, unless the defcustom `markdown-live-preview-export` is set to `'delete-on-export`. This is a bug; it should be deleting this buffer every time.


jrblevin:
Thanks!

PR #47, opened 2015-12-24
add defcustom for list indent width

cosmicexplorer:
I prefer to have my markdown lists indented with two spaces. This commit creates a defcustom `markdown-list-indent-width` and integrates it into list promotion and demotion functions (defaulting to `4`, the current default number of spaces to indent).

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1055,6 +1055,12 @@ export by setting to 'delete-on-export, when quitting
   :group 'markdown
   :type 'symbol)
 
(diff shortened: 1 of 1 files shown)


jrblevin:
Merged—thanks! I also added a basic test case in df0c732.

cosmicexplorer:
Thanks! I'm still not too familiar with testing for elisp, so thanks for finishing that up. I'll try to include appropriate test cases for large changes like this next time.

PR #49, opened 2016-01-02
allow killing of markdown-live-preview-* buffer

cosmicexplorer:
fixes an issue where if the live preview's output buffer was killed, the
mode would signal an error about a killed buffer. when the output (eww) buffer
is killed, and the markdown source buffer is saved, the output buffer is
regenerated and displayed.

two things:
1. `markdown-live-preview-switch-to-output` is also added as an interactive function which switches to/regenerates the output buffer. If `markdown-live-preview-mode` is off, it turns it on. Is that what a user would expect, or should we just ignore it?
2. `markdown-live-preview-switch-to-output` currently has no keybinding because I didn't want to pollute the pool any more, especially with 2.1 on the horizon. It could very easily be added, though, along with a menu option.

Alternatively, we could just drop `markdown-live-preview-switch-to-output` and save ourselves the trouble. I just thought it'd be helpful. Since saving the source buffer does the same thing, it may be more appropriate to just use that.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -5127,9 +5127,11 @@ current filename, but with the extension removed and replaced with .html."
   (interactive)
   (browse-url-of-file (markdown-export)))
 
-(defvar markdown-live-preview-buffer nil
+(defvar-local markdown-live-preview-buffer nil
   "Buffer used to preview markdown output in `markdown-live-preview-export'.")
-(make-variable-buffer-local 'markdown-live-preview-buffer)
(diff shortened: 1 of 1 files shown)


jrblevin:
Merged—thanks. I agree with your choices on both 1 (switching on live preview) and 2 (no keybinding for now).

cosmicexplorer:
I'm actually adding a few tests for this and other elements of `live-preview-*`; gimme an hour or less.

jrblevin:
Sounds good!

PR #50, opened 2016-01-02
add testing for live-preview-*, and fixes

cosmicexplorer:
After testing `live-preview-*` functions, I found a few silly things I was doing, like `markdown-live-preview-export` modifying `(current-buffer)`. These have been fixed now, so this pull request does three things (in order of the commits):

1. Add indentation properties to `defmacro` forms in the test suite so they can be automatically indented (this does not affect anything important at all), and add an `unwind-protect` around one of the test macros so temporary files get cleaned up.
2. Actually add tests for the live-preview functionality, including the feature which was just added in #49.
3. Fixes an issue with `display-buffer` not displaying a buffer in another window when it really should. This commit also affects the defun `markdown-other-window`.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -5142,6 +5142,7 @@ buffer. Inverse of `markdown-live-preview-buffer'.")
 
 (defun markdown-live-preview-window-eww (file)
   "A `markdown-live-preview-window-function' for previewing with eww."
+  (require 'eww)
   (eww-open-file file)
   (get-buffer "*eww*"))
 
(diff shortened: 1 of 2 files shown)


cosmicexplorer:
Hm, it couldn't find `eww-open-file` on the build server, and then it said it couldn't `(require 'eww)`. I'm checking out emacs 24.3 right now; I thought `eww` was bundled with it.

syohex:
`eww-open-file` is an autoloaded function. So `(require 'eww)` is not necessary.

cosmicexplorer:
Right, that would make sense. I just added some advice to check for the presence of `eww`, I'm trying it on 24.3 now.

cosmicexplorer:
Awesome, looks like it checks out now.

PR #51, opened 2016-01-02
Fix for older Emacs

syohex:
defvar-local was introduced at Emacs 24.3.


cosmicexplorer:
Oops, I'll keep that in mind next time.

jrblevin:
Thanks. I fixed a small conflict with recent additions and applied this as 2c2abec.

PR #53, opened 2016-01-03
Fix/cleanup eww live export

cosmicexplorer:
It makes more sense to check whether `'eww-open-file` is actually bound than all of `'eww`. Also, I left some test code in.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -5146,10 +5146,12 @@ buffer. Inverse of `markdown-live-preview-buffer'.")
 
 (defun markdown-live-preview-window-eww (file)
   "A `markdown-live-preview-window-function' for previewing with eww."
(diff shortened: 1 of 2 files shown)


cosmicexplorer:
Not sure why the the CI is failing; if you look at the details, it appears to be some issue with cask? I'm not 100% sure. It seems to work on my local copy of 24.3. Not sure what to do here, there doesn't seem to be a way to restart the CI.

cosmicexplorer:
Ok, looks like this passed now. The failures earlier were weird, though.

jrblevin:
Looks like it may have been a fluke related to pulling down Cask anew each time. It looks good to me now also.

jrblevin:
`master` was one commit ahead, so I squashed these and rebased, applying as commit c0bb1c9.

PR #54, opened 2016-01-03
completing-read gfm code languages

### LINKED ISSUE #38: Ability to disable markdown-insert-gfm-code-block prompt
I appreciate the option to specify a language name for gfm triple backticks pre-formatted code blocks, but 99% of the time, I always want to skip it.

Could we introduce a configurable variable for whether the prompt is raised, or nil, or some preferred default language name is used?

cosmicexplorer:
Fixes #38.

Does a few other things (although I swept the truly extraneous stuff to #53):

1. `defcustom`s for additional languages for gfm to complete outside of the given set, and also one to toggle whether or not to use `electric-backquote` at all.
2. Flattens the `markdown-regex-gfm-code-block-open/close` into a single regex, and uses that for highlighting. It was impossible to detect which line was supposed to contain a valid language otherwise.
3. Parses the current buffer for languages that aren't in the corpus and adds them to a buffer-local list, which is part of the `completing-read` input.
4. Consistently uses `markdown-replace-regexp-in-string`.

Concerns:
I use helm, so it's always somewhat slower to load than the default, but 335 candidates is proving to be pretty slow to bring up (and it's not the `cl-find` call, that only takes about .05 seconds once byte-compiled according to `(benchmark)`). So I was thinking many of the languages in linguist are pretty nonimportant, and I tried selecting about 100 of them which had names I could recognize, but that didn't seem to improve it for me at all, so I think it's just that helm's `completing-read` is slow. Let me know if you have those performance concerns and we can probably just triage the languages (although it'd have to be a pretty heavy triage since 100 candidates is still taking just over a third of a second to bring up the completions buffer).


jrblevin:
Fantastic—thanks for working on this. I don't have any performance issues on my end (nor do I use helm). Looks great to me so far. I will test it a bit more before merging.

jrblevin:
One small thing I noticed regarding capitalization: For the first fenced code block in a file, the capitalization in the pre-specified language list seems to dominate what is typed by the user. For example, if I type "Fortran" as the language name the empty code block will be inserted as

    ``` FORTRAN

    ```

while I would have expected

    ``` Fortran

    ```

I won't have time to poke around the code until later today at the earliest, but I'll take a look.

cosmicexplorer:
Yeah, check out my response in #59.

blaenk:
For what it's worth, I didn't really notice any performance issues and I use helm as well.

cosmicexplorer:
Cool! Guess my laptop is just slow.

PR #56, opened 2016-01-04
add async export

cosmicexplorer:
This is more of a WIP than a completed feature; it needs a bit (a lot) more documentation/testing (any testing)/refactoring. Basically, it adds a defcustom `markdown-export-async`, which, if non-nil, performs the export operation in an asynchronous subprocess. This is nice because it pauses a lot less when the user saves while working, especially for larger files.

This is done without duplicating all the export code for sync and async by replacing all calls to `markdown` (and all functions that call those functions which call `markdown`, etc) with use of the `markdown-do-sync-or-async` macro, which performs the normal synchronous operation if `asyncp` is nil, but creates a lambda which attaches to the process sentinel otherwise (which is the reason for the `lexical-binding: t;`, or sets its result to `markdown-async-result` if the process finished.

Once the `lexical-binding: t;` was added, I worked to fix some of the byte compiler errors. I would recommend taking a look at b5eb8cb closely; I did my best to fix what was obvious but I may have done something silly due to lack of familiarity with the code.

The default callback for the async export (set in `markdown-make-process-sentinel`) is to message the user saying it completed; I'm considering making that a defcustom as well, or just a boolean for whether or not to message the user at all.

I'll be working on this more, and I know I already have a pr up, so don't worry about getting to this super fast.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1353,7 +1353,7 @@ Function is called repeatedly until it returns nil. For details, see
   (save-excursion
     (goto-char start)
     (let ((levels (markdown-calculate-list-levels))
-          indent pre-regexp close-regexp open close stop)
+          indent pre-regexp close-regexp open close)
       (while (and (< (point) end) (not close))
         ;; Search for a region with sufficient indentation
         (if (null levels)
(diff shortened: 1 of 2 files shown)


cosmicexplorer:
Ok, how should this be dealt with then? @syohex the build fails on this. Should I suppress warnings, or (require 'cl) at runtime? I don't like that, but I'm not sure what else to do.

syohex:
> Should I suppress warnings, or (require 'cl) at runtime? I don't like that, but I'm not sure what else to do.

You should remove `eval-when-compile` around `(require 'cl)`. Because changes of this PR contains `gensym` which is cl function. Or use cl-lib functions and macros instead of cl.el(This PR enables lexical-binding and lexical-binding was introduced at Emacs 24. package.el was also bundled since Emacs 24. So it is no problem to use cl-lib.el. Emacs 24.1 and 24.2 users can download cl-lib by package.el)

cosmicexplorer:
Cool, we've just changed it all to use cl-lib.

jrblevin:
I just saw this and wondered if it might be of interest: https://github.com/ancane/markdown-preview-mode

cosmicexplorer:
I like that a lot, actually. Thanks for pointing it out. I thought I'd looked around for stuff like this before I wrote it, glad there's some existing code.

cosmicexplorer:
Superseded by #69.

PR #58, opened 2016-01-04
use cl-lib

cosmicexplorer:
I was having issues with cl vs cl-lib byte compiling correctly, so I just turned everything to cl-lib. If there's a reason we're sticking to cl then just close this.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -31,6 +31,7 @@
 ;; Maintainer: Jason R. Blevins <jrblevin@sdf.org>
 ;; Created: May 24, 2007
 ;; Version: 2.0
+;; Package-Requires: ((cl-lib "0.5"))
 ;; Keywords: Markdown, GitHub Flavored Markdown, itex
 ;; URL: http://jblevins.org/projects/markdown-mode/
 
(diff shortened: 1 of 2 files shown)


jrblevin:
Since cl-lib is now recommended, I agree that we should switch to that. I also don't personally have any reservations about requiring it at runtime since it has been bundled with Emacs for a while. Aiming for compatibility with Emacs before 24.3 or so is getting harder and harder. Perhaps we should add a note about the dependency for Emacs 24.1 and 24.2 users, who can use package.el?

cosmicexplorer:
A note about the dependency should definitely be added. I'd just stick it in the README under "Installation," since MELPA downloads dependencies automatically.

jrblevin:
Thanks. I went ahead and added it last night in 4260e68, but I put it under a new section titled "Compatibility".

cosmicexplorer:
Oops, just saw you added it (sorry, laptop just crashed). I'd recommend at least referencing the "Compatibility" section from "Installation", since the only people reading "Installation" are the ones not using MELPA, and therefore the ones that might need to know beforehand to install cl-lib.

jrblevin:
No worries. It was easy to miss. Here's the newly updated version: 1ee52a1. Let me know if you think it needs any other changes.

cosmicexplorer:
Yeah, that sounds good.

PR #60, opened 2016-01-05
fix #59

### LINKED ISSUE #59: User-preferred case in GFM language names
For the first GFM fenced code block inserted in a file, the capitalization in the pre-specified language list seems to dominate what is typed by the user. For example, if one enters "Fortran" as the language name the empty code block will be inserted as

    ``` FORTRAN

    ```

while one would have expected to instead see

    ``` Fortran

    ```

(This was previously discussed in #54, but since that PR was merged I am just opening a new issue as a reminder.)

cosmicexplorer:
If a user enters a language using `markdown-insert-gfm-code-block`, or it exists in the buffer when `markdown-parse-gfm-buffer-for-languages` is called, and the language has different capitalization than one already in the corpus, it will prompt for confirmation (if done interactively), and then add it to the buffer-local corpus, and insert the language literally as the user gave it.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -3285,18 +3285,8 @@ already in `markdown-gfm-recognized-languages' or
          (find-result
           (cl-find cleaned-lang (markdown-gfm-get-corpus)
                    :test #'equal)))
-    (if find-result (setq markdown-gfm-last-used-language find-result)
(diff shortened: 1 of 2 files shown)


jrblevin:
Works for me. Thanks again!

PR #63, opened 2016-01-05
don't add global hooks for live-preview

cosmicexplorer:
Take global hooks and make them buffer-local. `markdown-live-preview-remove-on-kill` applies to both the markdown buffers and the buffers used to preview the export, so it needs to be re-added in two places, where the export-viewing buffer is created, and when `markdown-mode` is turned on.

The test `test-markdown-ext/live-preview-exports` adequately covers whether these hooks run.


jrblevin:
Thanks. Applied as 002680d.

PR #64, opened 2016-01-06
fix case when no language given for suggesting gfm code block langs

cosmicexplorer:
Three things:

1. Previously, if no language was given for a gfm code block (so just three backticks open/close), `markdown-gfm-parse-buffer-for-languages` would error out because the `(match-string-no-properties 2)` call would return nil. This fixes that.
2. Adds a `?` to the second newline in a block, allowing empty code blocks (which github allows).
3. Allows using <kbd>C-g</kbd> to quit out of the `completing-read` for a language, and returns an empty string, so you can create code blocks with no specified language without having to backspace.


jrblevin:
Done--thanks!

PR #65, opened 2016-01-08
Safer convert to filename

csrhodes:
With sufficiently evil contents in [[Wikilinks]], it's possible to confuse emacs and tramp to the point that they won't let you switch away from the buffer.  Handle conditions appropriately when examining the filesystem for
files named by wikilink content.


jrblevin:
Thanks for these patches!

Issue #67, opened 2016-01-09
multiline italics face when not wanted


hmelman:
I'm using Package-Version: 20160109.957 with the mac homebrew emacs-24.5-mac-5.15 (which is  24.5.1).

I have a list as follows:
- something about function foo_bar
- something else about foo_bar

Everything between the two _ is in markdown-italic-face. I don't think it should be. Marked 2 doesn't display it as italics and neither does GitHub.

markdown-italic-underscore is nil.

jrblevin:
This is indeed a bug. Thanks for the report. For now, I've added a test case in 8eca1ff (which is expected to fail until this is fixed).

jrblevin:
OK, I believe this is fixed now. (See the discussion in PR #68 for additional details.)

PR #68, opened 2016-01-10
Fix issue67

syohex:
Highlight italic only following cases
- Italic start and end point are same line
- If italic start and end point are different line, both lines are not list

#### Original

orig

#### Fixed version

fixed

This is related to #67.
CC: @hmelman

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -2433,6 +2433,15 @@ Return nil otherwise."
                           (match-beginning 5) (match-end 5)))
     (goto-char (1+ (match-end 0)))))
 
+(defun markdown-list-p (pos)
+  (save-excursion
+    (goto-char pos)
+    (goto-char (line-beginning-position))
+    (looking-at-p markdown-regex-list)))
+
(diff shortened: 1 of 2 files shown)


jrblevin:
Thanks! I'll merge this, but I think there are some cases that might still slip through. For example, what about this:

```
* something about function
  foo_bar
* something else about
  foo_bar
```

I think checking all lines between `begin` and `end` for list markers will cover this case as well. Given that, might a regexp search for the list item regexp in the potential italic region be more efficient?

jrblevin:
Also, regarding the tests you added, with this one Markdown.pl *does* generate a list item with italics:

```
* foo_bar
foo_bar
```

Now I'm thinking that checking for `markdown-list-face` in the possible italic range is a safer approach. It also covers the hanging list example above and is more efficient since we've already done the hard work of parsing the nested lists.

jrblevin:
I updated this by adding a new test in 6aafece and changing to check for `markdown-list-face` in 84d94aa. Let me know if you see any problems.

PR #73, opened 2016-01-11
fix #71

### LINKED ISSUE #71: option to use lowercase language name in gfm blocks
First of all, woah!! Thanks a lot @cosmicexplorer for your great work on #54!

It looks and seems to work beautifully! I had been following what seemed to me like a convention which was to use the lowercase representation of language names. The implementation of #54 seems to use the capitalized language, e.g. `JavaScript` or `Python` versus `javascript` or `python`. I know it's such a minor gripe! But I was wondering if you could possibly add an option to perhaps `(downcase chosen-candidate)` before inserting it?

I already have many codeblocks that use the lowercase names (which I really prefer) and everyone else I've seen does the same thing, so I'm not too eager to break with that convention and have mixed styles.

cosmicexplorer:


blaenk:
Great!

jrblevin:
Applied as cf1ed9f. Thanks!

Issue #75, opened 2016-01-12
Markdown-mode navigation and cycle ignore R code chunk


walmes:
Hi @jrblevin,

I use markdown-mode all time. This Emacs mode is fantastic.
However, most part of time I'm working with code chunks of R language. In R, comments are preceded by hashs like markdown sections. This is the problem, because I like cycle (S-TAB) and navigation across sections `(C-c C-{n p u d})` but R comments inside code chunks are also visited. This code chunk can have a lot of comments (because are class tutorials). I would like ignore text inside chunks to make navigation more precise and fast. I tried use `#'`, another valid comment prefix in R, but markdown mode still got it. I tried also indent my comments by one space, bun didn't work to. A typical code chunk is

    ```{r}
    # Plot data.
    plot(dist ~ speed, data = cars)
    ```

When mark is inside this chunk, Emacs is at ess-mode. Maybe this can be used to narrow S-TAB and navigation imposing to do only for lines at markdown-mode. Do you have any idea how can ignore the code chunks?

Thanks.

vyp:
Also see #33, which I think is similar.

jrblevin:
Thank you. Yes, I thought this should have been fixed in 3192c9b, which was in response to #33. Can you confirm that you are using either the latest stable (2.1) or development (Git) version of markdown-mode?

What are you using for mode-switching? polymode or mmm-mode, maybe? Does cycling work As expected with that disabled?

syohex:
I suppose markdown-cycle issue is not fixed yet. How about this patch(https://gist.github.com/syohex/4060513d8bd7601c6afc), @walmes ?

### Original

![orig](https://cloud.githubusercontent.com/assets/554281/12260832/fc6e9bc0-b960-11e5-9973-3cc8a2abf4c8.gif)

### Applied above patch

![fixed](https://cloud.githubusercontent.com/assets/554281/12260835/02c6e4d2-b961-11e5-9723-19909aa03934.gif)

walmes:
Yes. I agree with you @syohex. I experienced the same thing in the cycling and navigation. See the gifs below (sorry for the Portuguese, but this one of the tutorials I'm work on).

![markdown-mode-cycle](https://cloud.githubusercontent.com/assets/1236798/12266483/8fdb911a-b92a-11e5-880f-45a00567d41e.gif)

[...]

PR #77, opened 2016-01-15
Fix moving same level header over code block issue

syohex:
This is related to #75.
CC: @walmes, @jrblevin

Test code is [here](https://github.com/jrblevin/markdown-mode/compare/master...syohex:moving-over-code-block?expand=1#diff-822bc997651d7df47039aa490c50fa8c).
@walmes Should I add more tests ? If I missed some test cases, please tell me them.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -4852,11 +4852,38 @@ Stop at the first and last headings of a superior heading."
   (interactive "p")
   (markdown-move-heading-common 'outline-forward-same-level arg))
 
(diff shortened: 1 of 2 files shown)


jrblevin:
This is interesting. There was already a test like this one. In fact, there is a special Markdown file `outline-code.text` designed to test this. It's strange that those were passing but this new test fails (before your patch).

jrblevin:
I merged this just now. Thanks once again!

PR #78, opened 2016-01-16
Make safe for avoiding infinite loop

syohex:
If the position is not changed by `move-fn`, it causes infinite loop. So the position is not changed then terminating loop.


jrblevin:
Merged--thank you!

Issue #72, opened 2016-01-11
scripts/get-recognized-gfm-languages.el skips languages with spaces


blaenk:
For example, `Emacs Lisp:`. This is because the regular expression matches on those not containing spaces.

jrblevin:
Thanks for reporting this. Since I merged the patches from @syohex in #82, I think this is fixed now and will close the issue.

PR #82, opened 2016-01-21
Fix language correcting script

syohex:
- Original code does not return language list. It returns value of kill-buffer
- Original code ignores language which contains space like 'Emacs Lisp'
- Replace space with '-' like 'Emacs Lisp' -> 'Emacs-Lisp'

And update language list.

This is related to #72.
CC: @blaenk


blaenk:
Good work!

jrblevin:
Great, thanks! I only have one question: Does GitHub expect or require the spaces to be replaced by hyphens? In my limited testing, it seems so. For example, blocks beginning with ```` ``` Emacs-Lisp```` are highlighted but those beginning with  ```` ``` Emacs Lisp```` are not. I can't find this documented anywhere though...

blaenk:
[This](https://github.com/github/linguist/blob/master/test/test_language.rb#L27-L29) seems to show that hyphens are indeed used.

jrblevin:
Thanks for finding that! I think the patch is probably good as is then.

After experimenting some more (in comments), it seems that to use spaces you have to enclose the language name in braces, like this ```` ``` {Emacs Lisp}````.

PR #87, opened 2016-02-01
Add MELPA badges

syohex:
You can see badges [here](https://github.com/syohex/markdown-mode/blob/4e26429fe94467c48323ca1a32e7db341280a84e/README.md#emacs-markdown-mode--)


jrblevin:
Thanks! I rebased (just to keep the history linear) and applied this as a032ba3. I also updated the `webpage.sh` script, which automatically generates the README.md file, in commit 3e87487.

PR #89, opened 2016-02-03
Don't insert empty titles

lunaryorn:
Guard against `title` being an empty string.  There's no use in inserting empty titles, or did I miss something?

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -2909,7 +2909,7 @@ be used to populate the title attribute when converted to XHTML."
         (insert url)
       ;; When no URL is given, leave cursor at END following the colon
       (setq end (point)))
-    (when (and title (> (length title) 0))
+    (when (> (length title) 0)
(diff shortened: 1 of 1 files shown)


markdown-mode.el
```diff
@@ -2909,7 +2909,7 @@ be used to populate the title attribute when converted to XHTML."
         (insert url)
       ;; When no URL is given, leave cursor at END following the colon
       (setq end (point)))
-    (when title
+    (when (and title (> (length title) 0))
```
syohex: Is `(> (length title) 0)` enough ? Because `(length nil)` returns 0.

lunaryorn: Oh, indeed, didn't think of this.

lunaryorn: @syohex I've simplified the check to `(> (length title) 0)`

jrblevin:
Thanks!

PR #90, opened 2016-02-07
add outline-minor-mode with markdown headings

cosmicexplorer:
`outline-minor-mode` allows traversing and folding headings, just like the `outline-mode` major mode. The keybindings don't conflict with any bindings already in `markdown-mode`, and allow folding just like `org-mode`. I couldn't think of what tests would be appropriate since this just adds the functionality `outline-minor-mode` already does, but I can definitely add some as required.

To play with it, just open a markdown file and try <kbd>C-c @ C-c</kbd> to hide, and <kbd>C-c @ C-e</kbd> to show a heading. There are more complex controls too, all the same as in `outline-mode`.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -6072,7 +6072,9 @@ before regenerating font-lock rules for extensions."
   (add-hook 'kill-buffer-hook #'markdown-live-preview-remove-on-kill t t)
 
   ;; do the initial link fontification
-  (markdown-fontify-buffer-wiki-links))
+  (markdown-fontify-buffer-wiki-links)
+  ;; add `outline-minor-mode' for free section hiding
+  (outline-minor-mode 1))
 
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.markdown\\'" . markdown-mode))


syohex:
`outline-regexp` is already set [here](https://github.com/cosmicexplorer/markdown-mode/blob/feat/folding/markdown-mode.el#L6026-L6027)

cosmicexplorer:
Oops, thanks. Removed.

jrblevin:
markdown-mode already has section folding functions which are robust to certain cases that cause outline-minor-mode to fail. Since it is based on a regular expression, it cannot distinguish between hash marks that are comments in code blocks and those that are actual ATX headings, so we run into issues like #33 and #75.

Is the main goal to use the well-known keybindings? If so, we can just add additional bindings that use the custom functions (`markdown-show-subtree`, `markdown-hide-subtree`, etc.).

cosmicexplorer:
Sorry, I didn't realize this had already been considered. Was looking for functions with the word "fold" in them and didn't find the ones already there. I'll look more thoroughly next time.

cosmicexplorer:
Although, what is the purpose of setting `outline-regexp` if `outline-minor-mode` isn't turned on? Is it so that people can add `outline-minor-mode` or something themselves?

jrblevin:
I think that was the original idea (allowing people to turn on `outline-minor-mode` if/when they wanted). At this point, I'd say we only use the name `outline-regexp` for historical reasons. I wouldn't really recommend using `outline-minor-mode` over the functions we have now (for both movement and folding).

cosmicexplorer:
Ok, makes sense.

PR #91, opened 2016-02-09
Add yaml metadata parsing

### LINKED ISSUE #66: Feature suggestion: Support GFM YAML metadata block
Some Markdown implementations, e.g., GFM and Pandoc, support YAML metadata blocks. For GFM (according to testing, since it seems undocumented) the YAML block must start at the beginning at the document (even blank lines before it are not allowed) and be wrapped in `---` (*exactly* three dashes) at the opening and closing, so an example document would be

```markdown
---
title: README
author: Zhiming Wang
date: January 8, 2016
---

This is an example GFM document with a YAML metadata block.
```

and this will be rendered like

screen shot 2016-01-08 at 12 21 10 pm

Note that the blank line after the closing `---` line is optional.

(Pandoc, on the other hand, has very lax rules for YAML metadata blocks, which allows them to be placed anywhere in the document, and allows closing with `...` instead of `---`. But Pandoc Markdown doesn't need to be supported.)

Currently, the line immediately above the closing `---` is interpreted as a setext header, which looks rather weird (especially if you have a real header):

screen shot 2016-01-08 at 12 22 42 pm

Would you please consider supporting the metadata block? I suppose the block itself could be highlighted just like a fenced code block, and I suppose it shouldn't be too hard to adapt the parsing code of fenced code blocks to this particular block. Thanks.

cosmicexplorer:
Also add testing and allow multiple pandoc metadata blocks. Fix #66.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1083,6 +1083,12 @@ completion."
   :group 'markdown
   :type 'boolean)
 
+(defcustom markdown-use-pandoc-style-yaml-metadata nil
+  "When non-nil, allow yaml metadata anywhere in the document, and allow
+ellipses to close a metadata region."
+  :group 'markdown
+  :type 'boolean)
+
 (defcustom markdown-live-preview-window-function
   'markdown-live-preview-window-eww
   "Function to display preview of Markdown output within Emacs. Function must
@@ -1374,7 +1380,7 @@ Group 3 matches the mathematical expression contained within.")
 (defconst markdown-regex-math-display
   "^\\(\\\\\\[\\)\\(\\(?:.\\|\n\\)*\\)?\\(\\\\\\]\\)$"
   "Regular expression for itex \[..\] display mode expressions.
-Groups 1 and 3 matche the opening and closing delimiters.
+Groups 1 and 3 match the opening and closing delimiters.
 Group 2 matches the mathematical expression contained within.")
 
 (defconst markdown-regex-multimarkdown-metadata
@@ -1382,8 +1388,16 @@ Group 2 matches the mathematical expression contained within.")
   "Regular expression for matching MultiMarkdown metadata.")
 
 (defconst markdown-regex-pandoc-metadata
-  "^\\(%\\)\\([ \t]*\\)\\(.*\\)$"
+  "^\\(%\\)\\([ \t]*\\)\\(.*\\(?:\n[ \t]+.*\\)*\\)"
   "Regular expression for matching Pandoc metadata.")
+
+(defconst markdown-regex-yaml-metadata-border
+  "\\(\\-\\{3\\}\\)$"
+  "Regular expression for matching yaml metadata.")
+
+(defconst markdown-regex-yaml-pandoc-metadata-end-border
+  "\\(\\.\\{3\\}\\|\\-\\{3\\}\\)$"
+  "Regular expression for matching yaml metadata end borders.")
 
 
 ;;; Syntax ====================================================================
@@ -1496,6 +1510,17 @@ Function is called repeatedly until it returns nil. For details, see
                          'markdown-blockquote
                          (match-data t)))))
 
+ (defun markdown-syntax-propertize-yaml-metadata (start end)
+  (save-excursion
+    (goto-char start)
+    (while (markdown-match-yaml-metadata end)
+      (put-text-property (match-beginning 1) (match-end 1)
+                         'markdown-metadata-key (match-data t))
+      (put-text-property (match-beginning 2) (match-end 2)
+                         'markdown-metadata-markup (match-data t))
+      (put-text-property (match-beginning 3) (match-end 3)
+                         'markdown-metadata-value (match-data t)))))
+
 (defun markdown-syntax-propertize-headings-generic (symbol regex start end)
   "Match headings of type SYMBOL with REGEX from START to END."
   (save-excursion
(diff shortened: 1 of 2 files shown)


markdown-mode.el
```diff
@@ -2599,30 +2633,60 @@ analysis."
         (set-match-data (list beg (point)))
         t))))
 
-(defun markdown-match-generic-metadata (regexp last)
+(defun markdown-get-match-boundaries (start-header end-header last &optional pos)
+  (save-excursion
+    (goto-char (if pos pos (point-min)))
```
syohex: How about `(or pos (point-min))` like other parts ?

markdown-mode.el
```diff
@@ -2599,30 +2633,60 @@ analysis."
...
+    (cl-loop
+     with cur-result = nil
+     while (and (< (point) last)
+                (re-search-forward (or start-header "\\`") last t)
+                (progn
+                  (setq cur-result (match-data))
+                  (re-search-forward
+                   (or end-header "\n\n\\|\n\\'\\|\\'") nil t)))
```
syohex: Following expressions are loop invariants. They can be moved outside the loop.
-  `(or start-header "\\`")`
- `(or end-header "\n\n\\|\n\\'\\|\\'")`

markdown-mode.el
```diff
@@ -2599,30 +2633,60 @@ analysis."
...
-                          (match-beginning 0)
-                        (point-max)))))
-    (cond ((>= (point) header-end)
-           ;; Don't match anything outside of the header.
+  ;; if start-header is nil, we assume metadata can only occur at the very top
+  ;; of a file ("\\`"). if end-header is nil, we assume it is "\n\n"
+  (let* ((header-bounds
+          (markdown-get-match-boundaries start-header end-header last))
+         (enclosing-header
+          (cl-first             ; just take first if multiple (maybe impossible)
```
syohex: Can `cl-first` + `cl-remove-if-not` be replaced with `cl-find-if` ?

tests/markdown-test.el
```diff
@@ -2270,7 +2271,8 @@ Base Header Level: 2  "
 
 (ert-deftest test-markdown-font-lock/mmd-metadata-after-header ()
   "Ensure that similar lines are not matched after the header."
-  (markdown-test-string "Title: peg-multimarkdown User's Guide  
+  (markdown-test-string
+   "Title: peg-multimarkdown User's Guide
```
syohex: Please do not change unrelated part.

tests/markdown-test.el
```diff
@@ -2282,7 +2284,8 @@ Base Header Level: 2  "
 
 (ert-deftest test-markdown-font-lock/pandoc-metadata ()
   "Basic Pandoc metadata tests."
-  (markdown-test-string "% title
+  (markdown-test-string
```
syohex: Ditto

tests/markdown-test.el
```diff
@@ -2252,7 +2252,8 @@ for (var i = 0; i < 10; i++) {
 
 (ert-deftest test-markdown-font-lock/mmd-metadata ()
   "Basic MultMarkdown metadata tests."
-  (markdown-test-string "Title: peg-multimarkdown User's Guide  
+  (markdown-test-string
+   "Title: peg-multimarkdown User's Guide
```
syohex: Please do not change unrelated part.

syohex:
Thanks for great work. I put some comments, please check them.

markdown-mode.el
cosmicexplorer: Fixed, not sure what I was thinking.

markdown-mode.el
cosmicexplorer: Also fixed.

markdown-mode.el
cosmicexplorer: Yes, that's what I was trying to do.

tests/markdown-test.el
cosmicexplorer: Removed.

syohex:
Looks good. Could you squash commits ?

cosmicexplorer:
Done.

syohex:
And please remove following lines from commit messages, they makes no sense.

> - Add defvar for markdown-search-until-condition
> - Update with comments from @syohex

cosmicexplorer:
Fixed.

syohex:
Thanks :+1:

jrblevin:
Thanks for this patch. I only noticed one potential issue: YAML blocks inside fenced code blocks should be ignored:

    ```
    ---
    title: foo
    author: bar
    ---
    ```

You can use `markdown-code-block-at-pos` or `markdown-code-block-at-point` in the propertize function to skip code blocks.

jrblevin:
Regarding my comment above about YAML in code blocks, I appended the test case and marked it with `  :expected-result :failed` for now. See commit 90f3f59. I took a shot at fixing it, but it's more complicated than I thought. I think we also need to check for code blocks in `markdown-get-match-boundaries`, because of the separate matching functions.

cosmicexplorer:
I'll take a look at these tomorrow. I think I can speed up the metadata block matching by propertizing metadata block start and end boundaries (the `---\n` lines) and searching for characters with that property. We can then search for code block boundaries as well, using the same trick.

With respect to the if statements, I wanted to allow for local binding of the pandoc style; if you want to remove the if statements, I can factor that out into a separate function.

vyp:
Thank you @cosmicexplorer for this! :+1:

cosmicexplorer:
Thinking about it a bit more, metadata would be invalid inside any other "block"-level construct, such as gfm code blocks or tilde blocks; I think that's it? All the other block constructs have some required characters at the beginning of each line. I think what would probably work would be to propertize all three "border" types (gfm blocks, pandoc tilde blocks, yaml metadata blocks) at once in a single pass, then figure out the appropriate context for any text enclosed within them by finding which if any characters with the syntax property for start/end blocks of those three block types are closest to the characters to propertize.

So basically, if our text is like:
```
... (some stuff without ~~~, ```, or ---)
`` ` (propertized as gfm-codeblock-begin by first block-construct pass)
--- (not propertized because already within block construct, but with markdown-inline-code-face)
key: value (look for enclosing block propertized text; found gfm-codeblock, so no propertizing/highlighting)
--- (not propertized)
`` ` (gfm-codeblock-end)
```

PR #97, opened 2016-02-18
use lexical binding and fix byte-compile warnings

cosmicexplorer:
This pull request adds `lexical-binding: t` to the top of the file and fixes byte-compile warnings. It's not necessary now, but it makes working with async processes in #69 easier.


syohex:
Thanks. Looks good.

PR #99, opened 2016-02-19
add make task to test all supported emacs with evm

cosmicexplorer:
This adds a task `evm` to `/tests/Makefile` to run the tests in a sub-make with evm binaries from `24.1` to `24.5`. I thought this might be nice to check test regressions on other emacs without waiting for travis to run on checkin to github. I'm not super familiar with EVM, maybe `24.*-travis` should be used instead of `24.*-bin`; I just assumed we shouldn't use travis since we're not running on a travis node. This also has some fixes which make it compile and run cleanly on all the versions being tested on.

This seems to be working fine, except for `emacs-24.1-bin`, which is segfaulting on running tests for reasons I don't understand (it's not happening in any specific tests, so maybe something nondeterministic like a memory error?). All other evm versions work just fine, so I was hoping you could shed some light on what the matter is here.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -2783,9 +2783,9 @@ GFM quoted code blocks.  Calls `markdown-code-block-at-pos'."
 
 ;;; Markdown Font Lock Matching Functions =====================================
 
-(defun markdown-range-property-any (begin end prop values)
-  "Return t if PROP from BEGIN to END is equal to one of the given VALUES.
-Also returns t if PROP is a list containing one of the VALUES.
+(defun markdown-range-property-any (begin end prop prop-values)
+  "Return t if PROP from BEGIN to END is equal to one of the given PROP-VALUES.
+Also returns t if PROP is a list containing one of the PROP-VALUES.
 Return nil otherwise."
   (let (props)
     (catch 'found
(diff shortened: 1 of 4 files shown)


cosmicexplorer:
Ah, the segfault actually appears to occur in `test-markdown-footnote-reference/jump`, only in evm `emacs-24.1-bin` on my arch linux. It might be because this is the only test that uses `execute-kbd-macro`? More investigation to come.

syohex:
EVM `-bin` packages are deprecated. EVM document says as below so that we should not use them.

> All -bin versions will are deprecated and will be removed. Do not use them!

cosmicexplorer:
Oops. Changed to just `emacs-24.x`.

syohex:
IMO, it is enough to use `make EMACS=$(evm bin emacs-24.1)` manually(or writing such shellscript) instead of this PR. This **make task** is not useful for all developers. EVM supports only MacOSX and Linux, other platform(Windows, BSD distributions etc) users cannot use this **make task**.

cosmicexplorer:
Yeah, I hadn't thought of that, and I agree, it's definitely easy to just write a shell script. Feel free to delete this then.

jrblevin:
Might the CL flags be useful in getting Travis tests working with older Emacsen? I couldn't get them working with versions prior to 24.3 previously.

cosmicexplorer:
Oh yeah, that plus the avoiding byte-compilation warnings should work (at least it works in my evm?). You'll have to actually have the cl-lib.el file in the repo, though.

jrblevin:
OK, let's just leave this open for now then. I'll see if I can use it to get the CI tests working on older Emacs versions.

jrblevin:
Done. See 3172552 and 915723d. We now have Travis CI testing with Emacs 24.1 through 24.5. Thanks!

cosmicexplorer:
Awesome, I'm glad we didn't find any lurking issues with earily emacs-24 versions.

Issue #98, opened 2016-02-19
Ordered list renumbering freeze


allencch:
When I have such content

```
a

1. a
3. c
2. b
```
And invoke the "C-c C-c n", Emacs will freeze.

But if I remove the empty line after "a" (the second line), then the renumbering works fine.

cosmicexplorer:
This happens to me too. I'm gonna look at this later today.

syohex:
@allencch Thanks for reporting issue. This issue is fixed by #100(Thanks @cosmicexplorer). Please check latest version. And please report us if you still have problem about this.

allencch:
@syohex @cosmicexplorer , thank you. It works fine now.

PR #100, opened 2016-02-19
fix cleanup-list-numbers infinite loop

cosmicexplorer:
This fixes the case given in #98. However, I couldn't reproduce #98 noninteractively; I'll look for a way to do that.


syohex:
I confirmed #98 issue is fixed with this PR.

>  However, I couldn't reproduce #98 noninteractively;

It looks `replace-match` behavior is different between interactive-mode and non-interactive-mode.

>  I'll look for a way to do that.

Thanks. We should add regression test about this.

PR #102, opened 2016-02-22
make live-preview follow min or max point

cosmicexplorer:
This makes the output buffer window's point for `markdown-live-preview-mode` stay at `point-min` or `point-max`, if the window displaying `markdown-live-preview-buffer` was already at `point-min` or `point-max`. This means the output buffer will not have to be manually re-scrolled down if the user is adding lots of text to to the bottom of a file and using live-preview.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -5872,21 +5872,63 @@ buffer. Inverse of `markdown-live-preview-buffer'.")
         (get-buffer "*eww*"))
     (error "eww is not present or not loaded on this version of emacs")))
 
+(defun markdown-visual-lines-between-points (beg end)
+  (save-excursion
+    (goto-char beg)
+    (cl-loop with count = 0
+             while (progn (end-of-visual-line)
+                          (and (< (point) end) (line-move-visual 1 t)))
+             do (cl-incf count)
+             finally return count)))
+
 (defun markdown-live-preview-window-serialize (buf)
   "Get window point and scroll data for all windows displaying BUF if BUF is
-non-nil."
-  (when buf
-    (mapcar (lambda (win) (list win (window-point win) (window-start win)))
-            (get-buffer-window-list buf))))
+live."
+  (when (buffer-live-p buf)
+    (with-current-buffer buf
+      (mapcar
+       (lambda (win)
+         (with-selected-window win
+           (let* ((start (window-start))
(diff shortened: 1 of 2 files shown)


cosmicexplorer:
Ok, for some reason this works just fine interactively, but behaves differently in batch mode (the eternal struggle), causing the test to fail. I'll find a way to trip this appropriately.

cosmicexplorer:
This might also fix some of the errors #69 is having too because of the fix to `markdown-temp-eww`. I also changed it to make sure the number of visual lines between `window-point` and `window-start` is used, instead of real lines.

Now that `make evm` exists, I'll fix the errors each version is having on these tests.

cosmicexplorer:
Ok, this should work. Most of the issue was that I forgot that `markdown-temp-eww` doesn't actually use `eww`, and I forgot that the CI server sometimes doesn't have `libxml2` compiled in. This should satisfy the CI server.

jrblevin:
For some reason `(featurep 'libxml2)` returns nil on my Emacs (25.1.50.1 on OS X), so the `test-markdown-ext/live-preview-exports` fails.

jrblevin:
The above mentioned 25.1 build was from [emacsforosx.com](http://emacsforosx.com). I tried a build of 24.5 from MacPorts and it also returns nil. Both have `eww`.

cosmicexplorer:
That's because `libxml2` isn't a "feature" in emacs lisp, I was going too fast and thought I remembered how libxml2 was checked. I've changed that to `(fboundp 'libxml-parse-html-region)`. That's weird that it was failing, though; it should just skip it. Can you describe what the error was? Either way, this should fix it?

jrblevin:
Thanks. That's right--I think we saw a similar issue before, but I forgot.

I'll check this version again soon. Here's the error I got before:

```
Test test-markdown-ext/live-preview-exports condition:
    (ert-test-failed
     ((should-error
       (markdown-live-preview-mode))
      :form
      (markdown-live-preview-mode)
      :value t :fail-reason "did not signal an error"))
   FAILED   22/248  test-markdown-ext/live-preview-exports
```

cosmicexplorer:
Oh ok, that makes sense. I was worried it was from somewhere else.

jrblevin:
Thanks, merged as dda54f7.

Issue #104, opened 2016-02-24
Can markdown-mode cause problems with query-replace?


chairraver:
Hi there.

I believe, I'm experiencing an issue with the Emacs query-replace function, which in the end might be caused by the latest development versions of markdown-mode from Melpa.

Let my try to explain. Let's say I've a simple Markdown file, which contains a simple bulleted list of links to a number of images in a directory 

    * [abcdeg](/xyz/pic[1-20].jpg)
    * [abcdeg](/xyz/pic[1-20].jpg)
    …

and the latest markdown-mode loaded (20160223.537 after having Emacs terminated and restarted). 

Then I execute "M-x query-replace" and enter "/xyz/" to be replaced by "/klm/". Then all occurrences of "/xyz/" are highlighted and I should be able to replace all occurrences one by one by using the Space key or all at once by hitting the ! key. What I'm experiencing however is, that the first occurrence is replaced and then query-replace basically stops and I'm back to editing. In the *message* buffer I see "replace-match-maybe-edit: Args out of range: 1336".

The reason, why I believe markdown-mode might have something to do with this behavior is, that I put the buffer into simple text-mode (M-x text-mode). Then query-replace worked as expected.

Additionally I also replaced the markdown-down.el file with the one from 2.1 and byte-compiled it and completely restarted Emacs (simple loading of the byte-compiled file does not seem enough). And with the 2.1 markdown-mode loaded query-replace also worked as expected.

Hopefully I don't pickup any residue from something else I've configure in my Emacs startup. I'm experiencing the described behavior on Windows and Linux Emacsen, which share cloud synced initialization files.

BTW, really thank you so much for a really great Emacs Major mode.

syohex:
Which emacs version do you use ?

chairraver:
GNU Emacs 24.5.1 (x86_64-w64-mingw32) of 2015-09-02 on WARLOCK

syohex:
I cannot reproduce this issue as below on Ubuntu 15.10 with Emacs 24.5.

![markdown-104](https://cloud.githubusercontent.com/assets/554281/13283554/ab01f488-db31-11e5-8716-d68273770898.gif)

Can you reproduce this issue by `emacs -Q -l markdown-mode.el` ?

chairraver:
Funny thing. It's there, then it isn't.

Since I'm currently @work I'm under Windows. Setup is:

[...]

PR #105, opened 2016-02-25
Save match data in markdown-syntax-propertize-extend-region

syohex:
This function should not overwrite match data.

I suppose this is related to following issues

- https://github.com/jrblevin/markdown-mode/issues/104
- https://github.com/syl20bnr/spacemacs/issues/5203

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1455,19 +1455,20 @@ This helps improve syntax analysis for block constructs.
 Returns a cons (NEW-START . NEW-END) or nil if no adjustment should be made.
 Function is called repeatedly until it returns nil. For details, see
 `syntax-propertize-extend-region-functions'."
-  (save-excursion
-    (let* ((new-start (progn (goto-char start)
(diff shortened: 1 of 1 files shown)


jrblevin:
Thanks for tracking this down. Were you able to reproduce #104? I wasn't, so I can't check whether this helps.

syohex:
> Were you able to reproduce #104?

No.

However I suppose we can confirm this with simple code.

```lisp
(defun test-command ()
  (interactive)
  (beginning-of-line)
  (when (re-search-forward "foo" nil t)
    (let ((beg0 (match-beginning 0)))
      (insert "A") ;; <- markdown-syntax-propertize-extend-region is called
      (cl-assert (= beg0 (match-beginning 0))))))
```

Open following markdown file

```
abcdefg

foo<- move cursor here and do M-x test-command
```

With original code, match data is overwritten and assertion is failed, while assertion is not failed with this patch.

jrblevin:
Thanks for confirming, and for the patch!

PR #107, opened 2016-02-26
unfinished: make markdown links fold into clickable links like org-mode

cosmicexplorer:
I've been trying to work on something like this for the past few days. Basically, org-mode turns all plaintext links into clickable links that go to elsewhere in the file, to places in other files, to web pages, whatever. I was looking at implementing something like that for markdown-mode which would basically just:
1. make the enclosing brackets and the url invisible
2. add text properties to the link text to make it a clickable link

This doesn't do either of those things yet, because to do that I would have had to be able to apply text properties to links. So I started off with emulating the current link highlighting with properties instead of font lock regexes. This currently doesn't work and hangs in weird situations; in particular, the text:

```
[a](a)a|
```

where `|` is the location of point, hangs when <kbd>RET</kbd> is pressed. In addition, `[]()` isn't highlighted correctly, although it _is_ propertized correctly. I don't know why that's the case, because `[a](a)` is highlighted and propertized correctly, and the regex allows for no characters in between brackets.

Would this be a feature that would be cool to finish and then enable with a toggle? Is there anything obvious I'm doing that might cause the current behavior to hang and incorrectly highlight?

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1166,7 +1166,7 @@ curly braces. They may be of arbitrary capitalization, though."
   "Regular expression matches HTML comment closing.")
 
 (defconst markdown-regex-link-inline
-  "\\(!\\)?\\(\\[\\)\\([^]^][^]]*\\|\\)\\(\\]\\)\\((\\)\\([^)]*?\\)\\(?:\\s-+\\(\"[^\"]*\"\\)\\)?\\()\\)"
+  "\\(!\\)?\\(\\[\\)\\([^]\n]*?\\)\\(\\]\\)\\((\\)\\([^)\n]*?\\)\\(?:\\s-+\\(\"[^\"\n]*?\"\\)\\)?\\()\\)"
   "Regular expression for a [text](file) or an image link ![text](file).
 Group 1 matches the leading exclamation point (optional).
 Group 2 matches the opening square bracket.
@@ -1178,7 +1178,7 @@ Group 7 matches the title (optional).
 Group 8 matches the closing parenthesis.")
 
 (defconst markdown-regex-link-reference
-  "\\(!\\)?\\(\\[\\)\\([^]^][^]]*\\|\\)\\(\\]\\)[ ]?\\(\\[\\)\\([^]]*?\\)\\(\\]\\)"
+  "\\(!\\)?\\(\\[\\)\\([^]\n]*?\\)\\(\\]\\)[ ]?\\(\\[\\)\\([^]\n]*?\\)\\(\\]\\)"
   "Regular expression for a reference link [text][id].
 Group 1 matches the leading exclamation point (optional).
 Group 2 matches the opening square bracket for the link text.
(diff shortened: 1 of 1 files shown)


jrblevin:
Sorry I haven't gotten back to you yet. I haven't seen org-mode's link folding yet, but I will take a look. I also have an incomplete local branch with a matching function for links and I have some ideas about how to best implement this, but it may be a couple of days before I can write them down in more detail. When investigating #110 I found some more general issues with the way cached match data is used that should be fixed first.

cosmicexplorer:
Ok, no worries. I'll take a look at that too when I can.

jrblevin:
I was looking for something else and found this. Maybe it is of interest?
http://codereview.stackexchange.com/questions/94275/in-emacs-markdown-mode-making-the-link-live-and-url-invisible-in-inline-links

cosmicexplorer:
That sounds really promising. My problem was that I was getting that weird
error when converting link fontification to use text properties instead of
a font lock regex, but this may sidestep that, which would be great to
avoid. Sorry for not stepping on this, classes have been eating up my time.
Thanks a lot for finding this, I'll see if I can adapt it. Do you know if
there's licensing issues? It looks like SO has creative commons by default,
but that might be changing soon (
http://meta.stackexchange.com/questions/272956/a-new-code-license-the-mit-this-time-with-attribution-required
).

jrblevin:
Yeah, sorry for the slow response on my end too. Still haven't looked at org-mode links, but I will. And I'll dig up my own branch to see what the hang up was there.

Honestly, I didn't look too closely at the code on SO. I'm not necessarily endorsing what's there, but given the topic I thought I'd pass it along in case there was anything useful. 

Briefly, my thoughts are that for inline elements like links we should skip the text property phase and jump straight to fontification with a matching function (and match all components in one pass, rather than making one pass for link text, another for the URL part, etc.).

cosmicexplorer:
Makes sense, both performance-wise and complexity-wise. I thought it'd make
it easier since invisible text requires setting properties anyway, but like
I said I'll check this out.

podiki:
Was there ever any progress on this? Folding of links would be a huge boost to readability in Emacs. I would definitely take inspiration from org-mode, as it handles links very nicely.

jrblevin:
I agree. It's high on my list now--long inline links have been bugging me a lot lately--and I have some time to work on it. I've got a local working branch, where I'm taking a different approach than this patch, but it's not ready yet.

As an aside, this is related to issue #130.

podiki:
Sounds good! I'm happy to test or provide any feedback in the future.

jrblevin:
Check out the latest master. Here are the relevant changes:

-   Make inline links, reference links, angle bracket URLs, and plain URLs clickable.
-   URL and reference label hiding for inline and reference links. This is configurable via `markdown-hidden-urls`.  By default, URLs will appear as `[link](∞)` instead of `[link](http://perhaps.a/very/long/url/)`.  To change the placeholder character used, set `markdown-url-compose-char`.  This feature can be toggled using <kbd>C-c C-x C-l</kbd> (`markdown-toggle-hidden-urls`).

The remaining part is to write an easy link editing function, but as in org-mode you can delete the last character to edit the URL for now.

podiki:
Excellent, thanks! Just took a quick look and it works fine on my end.

Any reason for displaying as `[link](∞)` rather than just `link` (or `[link]`)? Or is that just a placeholder until there is an link inserting and editing function as in org-mode, to easy editing links?

Thanks again!

jrblevin:
Great!

With Org, there is only one type of link (I think). In Markdown, however we have inline and reference links (and wiki links as an extension). The only way to tell the difference is if we keep the final brackets or parentheses, so I didn't hide those (yet). It's also part of a bigger plan, in which we will have an option to hide most or all markup (including the link brackets) as a separate level of hiding.

podiki:
Sounds good, was just curious. Definitely already a nice improvement to having long links everywhere.

PR #109, opened 2016-02-26
Respect narrowed region in markdown-find-previous-prop

vspinu:
Hi, 

In multi-modes like [`polymode`](https://github.com/vspinu/polymode) all `markdown-mode` functionality like font-locking and syntax properties is executed in narrowed regions. 

It would be great if the code would not rely on assumptions like buffer starting at 1. This is a fix of one of such issues, there might be others. I will report if I find more.


jrblevin:
Indeed, `(point-min)` is better than `1` here. Thanks for fixing this and please do report any other similar problems.

PR #111, opened 2016-03-02
Don't override match-data as possible

syohex:
- [x] Use `string-match-p` instead of `string-match` as possible
- [x] Use `looking-at-p` instead of `looking-at` as possible
- [x] Remove needless `save-match-data` block
- [x] Replace `markdown-cur-line-indent` with `current-indentation`

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1505,7 +1505,7 @@ region to refontify."
          ((and (looking-at pre-regexp)
                (save-match-data (markdown-prev-line-blank-p)))
           (setq open (match-beginning 0))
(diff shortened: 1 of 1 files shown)


jrblevin:
This looks great so far, thank you. I didn't know about the predicate (`-p`) versions of `looking-at` and `string-match`. That is extremely helpful to know.

Unfortunately, some of the markdown predicate functions likely do modify the match data, despite their names, so I will take a look at those.

syohex:
I found we can replace `markdown-cur-line-indent` with `current-indentation`

syohex:
Done

jrblevin:
Fantastic! Thank you.

PR #114, opened 2016-03-03
Add .dir-locals.el and fix all checkdoc errors

vspinu:
Given the trailing whitespace issue in last commit I am adding a dir-locals with a bunch meaningful settings. Also fixing all checkdoc errors.


jrblevin:
I forget to run checkdoc periodically, so instead I added a step to the automated tests that also runs checkdoc and reports errors (in dd7fd4c). Hopefully it will be easier to stay compliant going forward.

Issue #115, opened 2016-03-10
crazy buffer-file-name: nil, file-modified-p: t 


dabrahams:
Create a file containing the following:

```markdown
---
x: x
---
`x`
```

then open it in markdown-mode in emacs-24.5.1

For me, the buffer loses its relationship to a file and is marked as modified.

jrblevin:
Thanks for the report. I confirmed this on Emacs 24.5.1. Oddly, on Emacs 25.1 the buffer is still marked as modified but `buffer-file-name` remains intact.

jrblevin:
I think @syohex fixed this, but please reopen if the issue persists for some reason. Thanks again for reporting. This was a strange bug indeed.

PR #116, opened 2016-03-11
Correct beginning of inline code position

syohex:
markdown-regex-code starts with **"\\(?:\\`\\|[^\\]\\)\\(\\(`+\\)..."**. If cursor is beginning of buffer, `(match-beginning 0)` is on backtick. While cursor is not beginning of buffer, `(match-beginning 0)` is not on backtick and previous point of backtick.

```
---
foo:bar
---
`x`
```

In above example, `(match-beginning 0)` is point of after '---'. That point is a code block and '(markdown-code-block-at-pos pos)' returns non-nil, this causes infinite loop of markdown-match-inline-generic. `(match-beginning 1)` of markdown-regex-code always on backtick. We should use `(match-beginning 1)` instead of '(match-beginning 0') in such case.

This is related to #115.
CC: @dabrahams

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -2805,7 +2805,7 @@ Return nil otherwise."
 (defun markdown-match-inline-generic (regex last)
   "Match inline REGEX from the point to LAST."
   (when (re-search-forward regex last t)
-    (let ((bounds (markdown-code-block-at-pos (match-beginning 0))))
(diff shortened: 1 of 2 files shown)


syohex:
There is no error with this patch.

fix

jrblevin:
Thanks--nice catch!

PR #118, opened 2016-03-13
Remove autoload for .text

purcell:
While .md and .markdown are clearly markdown files, the handling of .text is not a choice that markdown-mode should be making on the user's behalf without any opportunity to prevent it.


jrblevin:
I'm inclined to apply this patch unless there are any well-reasoned objections. I'll wait a couple of days to see if there are any comments.

PR #124, opened 2016-03-19
Prevent spurious bold fontification

peniblec:
In markdown-match-bold, add guard clauses similar to those in
markdown-match-italic, to make sure bold markers in code or math do not
result in bold fontification, eg in cases such as:

* `**foo` bar `baz**`
* `def __init__(self):`


syohex:
Looks good to me.

syohex:
Sorry for late reply. Thanks for PR.

PR #125, opened 2016-03-25
Maintain cursor position when indenting

ihodes:
Now, when `markdown-indent-line' is called, the cursor stays on the
character it was on, instead of ending up at the beginning of the line.

Example: http://link.isaachodes.io/3Q04333E1j2k

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -4161,12 +4161,19 @@ these cases, indent to the default position.
 Positions are calculated by `markdown-calc-indents'."
   (interactive)
   (let ((positions (markdown-calc-indents))
+        (cursor-pos (current-column))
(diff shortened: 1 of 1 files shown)


markdown-mode.el
```diff
@@ -4161,12 +4161,19 @@ these cases, indent to the default position.
 Positions are calculated by `markdown-calc-indents'."
   (interactive)
   (let ((positions (markdown-calc-indents))
+        (cursor-pos (current-column))
+        (_ (back-to-indentation))
```
ihodes: This looks a little gross, but the alternative is another layer of let-bindings; let me know if you see a third path, or prefer something else!

PR #126, opened 2016-03-30
Remove redundant "Emacs" from package description

purcell:
- All elisp is for Emacs, so this is just noise in the packages list
- Removed 'files' too, since major modes are also for buffers without files


syohex:
Thanks

Issue #122, opened 2016-03-18
Add option to display images inline


ihodes:

syohex:
#128 have been merged.

PR #128, opened 2016-04-02
Implement inline image feature

syohex:
This is related to #122.
CC: @ihodes 

Should we add key bindings for those commands by default ?

See
- https://twitter.com/syohex/status/715544541915324416

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -6527,6 +6527,46 @@ BEG and END are the limits of scanned region."
        (progn (goto-char end) (forward-line 1) (point))))))
 
 
+;;; Display inline image =================================================
+
+(defvar markdown-inline-image-overlays nil)
+(make-variable-buffer-local 'markdown-inline-image-overlays)
+
+(defun markdown-remove-inline-images ()
+  (interactive)
+  (mapc 'delete-overlay markdown-inline-image-overlays)
+  (setq markdown-inline-image-overlays nil))
+
+(defun markdown-display-inline-images ()
+  (interactive)
+  (unless (display-graphic-p)
+    (error "Cannot show images."))
+  (save-excursion
+    (save-restriction
+      (widen)
+      (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


jrblevin:
Wow, only 40 lines. I thought this would be a much larger undertaking! Nice work. I won't be able to actually test it out for a day or two, but I looked at the patch and it looks good to me.

I think keybindings are a good idea. Image insertion is under <kbd>C-c C-i</kbd>, so maybe <kbd>C-c C-i C-t</kbd> for toggling?

ihodes:
That's awesome, thank you! I was also assuming it'd be a lot more code to get that to work.

markdown-mode.el
```diff
@@ -6527,6 +6527,46 @@ BEG and END are the limits of scanned region."
...
+  (mapc 'delete-overlay markdown-inline-image-overlays)
+  (setq markdown-inline-image-overlays nil))
+
+(defun markdown-display-inline-images ()
+  (interactive)
+  (unless (display-graphic-p)
+    (error "Cannot show images."))
+  (save-excursion
+    (save-restriction
+      (widen)
```
ihodes: Do you need to widen here? 

My question's context: I'm thinking for my use-case I'll probably narrow to region of the day I'm writing in, and only want the pictures toggles for that day (toggling them for all days could make things slow, particularly because I intend to hook into this code & use ImageMagik to make the photos no wider than the buffer size).

syohex: I took most of code from org-mode and org inline image commands uses `widen`. Commands which I add works same as org-mode commands(`org-toggle-inline-images` etc).

syohex:
> I think keybindings are a good idea. Image insertion is under C-c C-i, so maybe C-c C-i C-t for toggling?

@jrblevin Thanks for suggestion. I'll add bindings later.

syohex:
I have set toggle command to `C-c C-i C-t`.

PR #138, opened 2016-05-07
Remove needless code

syohex:
url-parse was bundled since Emacs 24.1 so that markdown-mode need not to
check whether it is installed. However url-path-and-query was introduced
at Emacs 24.3, so it should be checked.


jrblevin:
Looks good. Thanks for cleaning this up.

Issue #140, opened 2016-05-11
Possible wrong indication about direct download in README.md


deb0ch:
At the **direct download** section you suggest to write `(autoload 'gfm-mode "gfm-mode" ...` in our .emacs, but should not it be more like `(autoload 'gfm-mode "markdown-mode" ...` ?

It is possible that my understanding is wrong, but the first gave me a "file not found" error when opening .md files while the second worked just fine.

Cheers.

jrblevin:
Thanks, you're absolutely right: it should be `(autload function filename)`. I'll fix this soon.

syohex:
Done

PR #141, opened 2016-05-13
Fix wrong file name

syohex:
This is related to #140.


jrblevin:
Thanks!

HenryGessau:
The wrong file name is still showing on the project page, http://jblevins.org/projects/markdown-mode/

syohex:
I suppose that site is not updated. Please check README.md in github

jrblevin:
I updated the incorrect autoload filename on the homepage just now. Thanks for pointing that out.

(I do update that page, but only when a new release is made. That way the documentation there corresponds to the current release version rather than to things in development.)

PR #142, opened 2016-05-27
Don't return point which is larger than point-max

syohex:
Original code sometimes returns point larger than point-max, then
syntax-propertize function raises args-out-of-range signal.

This is related to
- http://qiita.com/PharaohKJ/items/a5b86404cfd60af68089(In Japanese)

CC: @PharaohKJ


PharaohKJ:
👍

syohex:
Thanks for checking.

Issue #145, opened 2016-06-22
frontmatter metadata showing up in imenu


blaenk:
I just noticed that similar to #136, for the following metadata:

```
---
title = "Blah"
comments = false
---
```

It shows `comments = false` as the first imenu entry.

jrblevin:
Thanks for the report. It should be fixed now, thanks to @syohex, but please reopen if not.

PR #154, opened 2016-07-21
Don't correct header like statement in metadata

syohex:
##### before

before

##### With this PR

after

This is related to #145
CC: @blaenk


syohex:
Tests on Emacs 24.5 are failed, however it looks network issue.

blaenk:
Thank you @syohex you are amazing!

jrblevin:
Thanks, @syohex!

Issue #166, opened 2016-09-19
Backspace doesn't delete region


cpitclaudel:
Markdown mode's redefinition of `<backspace>` doesn't handle the case in which a region is selected before pressing backspace.  In that case, the plain Emacs implementation deletes the full region, whereas markdown-mode only unindents.

To reproduce, create a new md file and insert three blank lines, followed by a line containing 8 spaces. Select all text, then press backspace.  The issue only exists when the last line contains whitespace.

syohex:
#167 have been merged. Could you check latest version ?

jrblevin:
Seems fixed to me, so I'm closing this.  Please reopen if not.  As always, thanks for the report!

PR #167, opened 2016-09-20
Backspace always deletes charactres if region is specified

syohex:
`markdown-exdent-or-delete` works only current line. So I suppose that it works as default backspace behaviour(deleting characters) if region is specified.

This is related to #166.
CC: @cpitclaudel


jrblevin:
Thank you, @syohex!

PR #168, opened 2016-09-23
use non-greedy *? for display math content

kleinschmidt:
Using greedy `*` means that the regex will match across display math blocks. For example, the current regex matches this whole snippet (including the non-math bit in the middle):

```
\[
some math
\]
non-math
\[
more math
\]
```

The only potential downside I can see is that if, for some strange reason, someone wants a literal `\]` inside their display math block, the regex will think that the block ends early. But this seems less likely than the frustrating situation of having non-math font-locked as math in between two display math blocks (which is very common in my documents).

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -6476,9 +6476,9 @@ This is an exact copy of `line-number-at-pos' for use in emacs21."
     ;; Update font lock keywords with extensions
     (setq markdown-mode-font-lock-keywords
           (append
+           (markdown-mode-font-lock-keywords-math)
            markdown-mode-font-lock-keywords-basic
-           (markdown-mode-font-lock-keywords-wiki-links)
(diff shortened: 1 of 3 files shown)


syohex:
Could you add test code ?

kleinschmidt:
I'll give it a try, but fair warning, my elisp skills are a little lacking :)

kleinschmidt:
Okay, tests added (and I confirmed they fail with greedy matching).

kleinschmidt:
I also added a test and a fix for a bug with italics and display math, where the font lock rules for display math were after the ones for italics and so the fences that normally keep `_` from being interpreted as italics in math mode weren't catching display math.

syohex:
Thanks for work. I'll check.

syohex:
Looks good.

PR #196, opened 2017-06-10
Fix #195

### LINKED ISSUE #195: `markdown-reload-extensions` should ensure fontification
## Expected behavior

Refontifying `markdown-mode` buffer should ensure that fontification completes.

## Actual behavior

Due to a lack of `font-lock-ensure` in `markdown-reload-extensions`, sometimes markdown buffers are not completely fontified before being displayed.

E.g.

image

Note that the beginning of the fence is not fontified.

This can be prevented by calling `font-lock-ensure` after `font-lock-refresh-defaults` in `markdown-reload-extensions`.

## Versions of `markdown-mode`, Emacs, OS, etc.

- `markdown-mode`: `20170609.942`
- Emacs: `25.1.1`
- OS: `Fedora 24`

xiongtx:
Ensure fontification before buffer is displayed.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -7413,7 +7413,8 @@ return the number of paragraphs left to move."
             (font-lock-syntactic-face-function . markdown-syntactic-face)))
     ;; Refontify buffer
     (when (and font-lock-mode (fboundp 'font-lock-refresh-defaults))
-      (font-lock-refresh-defaults))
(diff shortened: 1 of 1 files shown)


xiongtx:
Apparently `font-lock-ensure` was introduced in Emacs 25 and is not available in Emacs 24 and below. Oh well, it's not that important anyway.

PR #201, opened 2017-06-15
Don't error when previewing buffer without file

### LINKED ISSUE #200: `markdown-live-preview-mode` errors for buffers without file
## Expected behavior

`markdown-live-preview-mode` should gracefully fail for markdown buffers with no backing file.

## Actual behavior

`markdown-live-preview-mode` throws a `wrong-type-argument` error.

This is because `(markdown-live-preview-get-filename)` is nil for such buffers, which causes `markdown-live-preview-remove` to call `file-exists-p` on nil.

## Steps to reproduce (recipe)

1. Create a `markdown-mode` buffer without a backing file.
2. Call `markdown-live-preview-mode`

## Backtraces if necessary (`M-x toggle-debug-on-error`)

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  file-exists-p(nil)
  (if (file-exists-p outfile-name) (progn (delete-file outfile-name)))
  (let ((outfile-name (markdown-live-preview-get-filename))) (if (file-exists-p outfile-name) (progn (delete-file outfile-name))))
  (progn (let ((outfile-name (markdown-live-preview-get-filename))) (if (file-exists-p outfile-name) (progn (delete-file outfile-name)))))
  (if (eq markdown-live-preview-delete-export (quote delete-on-destroy)) (progn (let ((outfile-name (markdown-live-preview-get-filename))) (if (file-exists-p outfile-name) (progn (delete-file outfile-name))))))
  markdown-live-preview-remove()
  (if markdown-live-preview-mode (markdown-display-buffer-other-window (markdown-live-preview-export)) (markdown-live-preview-remove))
  ...  
```

## Describe versions of `markdown-mode`, Emacs, OS, etc.

- `markdown-mode`: `20170614.1152`
- Emacs: `25.1.1`
- OS: `Fedora 24`

xiongtx:
Fixes #200.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -7088,33 +7088,35 @@ Browse the resulting file within Emacs using
 `markdown-live-preview-window-function' Return the buffer
 displaying the rendered output."
   (interactive)
-  (let* ((markdown-live-preview-currently-exporting t)
-         (cur-buf (current-buffer))
-         (export-file (markdown-export (markdown-live-preview-get-filename)))
-         ;; get positions in all windows currently displaying output buffer
-         (window-data
-          (markdown-live-preview-window-serialize
-           markdown-live-preview-buffer)))
-    (save-window-excursion
-      (let ((output-buffer
-             (funcall markdown-live-preview-window-function export-file)))
-        (with-current-buffer output-buffer
-          (setq markdown-live-preview-source-buffer cur-buf)
-          (add-hook 'kill-buffer-hook
-                    #'markdown-live-preview-remove-on-kill t t))
+  (let ((filename (markdown-live-preview-get-filename)))
+    (when filename
+      (let* ((markdown-live-preview-currently-exporting t)
+             (cur-buf (current-buffer))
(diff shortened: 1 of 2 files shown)


jrblevin:
Thanks for this. It looks OK to me, but `user-error` is not defined on Emacs 24.1 and 24.2, so the CI tests are failing. Can you please fix that? Two possibilities are:

1. Maybe just use `error` instead?
2. Add a `markdown-user-error` `defalias` that falls back to `error` when `user-error` is not bound. See the "Compatibility" section of the code for some examples.  Something like this should work:

``` emacs-lisp
(eval-and-compile
  (defalias 'markdown-user-error
    (if (fboundp 'user-error)
        'user-error
      'error)))
```

xiongtx:
Yeah, I tried to use `skip-unless` but that actually caused the Emacs 24.3 test to fail as well.

In the end I just went with `error`. At some point support for Emacs <24.3 should probably be dropped.

PR #214, opened 2017-06-29
add contents feature

plandes:
This feature inserts a table of contents into the markdown file by looking for leading hashes `(#)` and using the header text.  Links are created for each header to each section found.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -8700,8 +8700,7 @@ If BUFFER is nil, use the current buffer"
 	   (-map #'(lambda (elt)
 		     (let ((text (cdr (assoc 'text elt)))
 			   (level (cdr (assoc 'level elt))))
-                       (->> (make-list (* 2 (- level 2)) " ")
-                            (apply #'concat))
+                       (insert (apply #'concat (make-list (* 2 (- level 2)) " ")))
 		       (insert (format "* [%s](#%s)\n" text
                                        (markdown-title-to-link text)))))))
       (buffer-string))))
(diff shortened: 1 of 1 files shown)


jrblevin:
Thanks for taking the time to submit this patch and fix it up.  Here are a few thoughts:

*   There is an existing [`markdown-toc`](https://github.com/ardumont/markdown-toc) package.  It has the added benefit of being able to update the existing table of contents when called again.  At a minimum, I think we should coordinate with that package author before adding this to `markdown-mode`, since that code seems fairly mature.

*   The regular expression used only matches atx headings.  There is an existing regular expression defined, `markdown-regex-header`, which also matches setext headings.

*   A regular expression approach will be fooled by hash marks inside code blocks (shebangs, etc. are very common).  You need a guard clause to avoid matching inside code blocks, such as `(unless (markdown-code-block-at-point-p) ...`

*   There is already a function which parses the heading tree: `markdown-imenu-create-nested-index` generates a table of contents for `imenu`.  You could pretty easily reuse that data structure I think.

*   As a benefit, reusing existing the functions will also avoid the `dash` dependency.

*   The `markdown-title-to-link` function needs to handle special characters in headings, and so on.  Following Pandoc and/or MultiMarkdown for this would be ideal.

plandes:
I have installed, customized and started to use markdown-toc.

You're right on all points and it appears that I reinvented a whee--an inferior wheel at that.

Please disregard this pull request.

plandes:
Also, thanks for your input.

jrblevin:
No problem—thanks for the PR.

PR #229, opened 2017-07-27
Add `markdown-add-gfm-checkbox` and make it a final fallback for `markdown-do`

knu:
I often just start writing a list item and notice it should have a checkbox, so I need something like this.


jrblevin:
Nice idea and thanks for sending a patch. I am slowing down adding features so that I can release v2.3, but I will get this in soon after that.

I appreciate that you added a test case as well—that really helps!

knu:
Could this be merged for 2.4? 😊

jrblevin:
Yes, thanks for the reminder. I committed it as 7772fc9 (I rebased it and also shortened the commit message so that it doesn't wrap). I made a couple of changes in subsequent commits for consistency with other similar commands: the new command is now named `markdown-insert-gfm-checkbox` and it is now at `C-c C-s [` (along with other insertion commands under `C-c C-s`). Thanks again for the patch!

PR #236, opened 2017-08-13
Add option markdown-gfm-uppsercase-checkbox

rnkn:
If non-nil, use [X] for completed checkboxes, [x] otherwise. Default nil.

Although I prefer the aesthetics of lowercase, GFM allows both, whereas Org only
allows uppercase. In the interest of cross-compatibility, this options allows
editing a checklist in either mode.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1347,6 +1347,11 @@ This applies to insertions done with
   :group 'markdown
   :type 'boolean)
 
+(defcustom markdown-gfm-uppercase-checkbox nil
+  "If non-nil, use [X] for completed checkboxes, [x] otherwise."
+  :group 'markdown
+  :type 'boolean)
+
 (defcustom markdown-hide-urls nil
   "Hide URLs of inline links and reference tags of reference links.
 Such URLs will be replaced by a single customizable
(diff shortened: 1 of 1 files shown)


rnkn:
I'm not sure how to make tests sorry...

jrblevin:
Thanks! I committed this as a007a494f150c20efc924c3c994de81ad30b7697 (where I fixed a typo in the commit message). I also added some documentation in 89d63a66b1fd9e353c0047917c0c44856cbd922c.

rnkn:
Awesome.

Can you help me out... so my repository history now diverges from yours... just to keep my side tidy if I wanna contribute something in the future, what I've done is pulled from yours, then force pushed back to mine... is this the right thing to do?

jrblevin:
That's exactly what I would have done, and your history looks the same as mine, so I think you should be good now.

In the future, another option is to make a new "feature branch" and issue the PR from that branch. Then if we happen to diverge you can just delete that feature branch and pull from `master` to get back in sync. It saves maybe one or two steps, but if you're comfortable occasionally doing a hard reset and force push it really doesn't matter.

rnkn:
Cool, thanks.

PR #252, opened 2017-09-04
Fix markdown-inline-code-face's :inherit attribute

peniblec:
Hi! When inline code lost its foreground color some months ago,
I assumed it was a stylistic choice and did not question it; when
I tried to Customize the face recently, however, the Custom menu
spewed a broken face form.

AFAIK :inherit takes either a single face, or a sequence of faces, so
I guess the declaration ought to be fixed?

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -2515,7 +2515,7 @@ See `markdown-hide-markup' for additional details."
   :group 'markdown-faces)
 
 (defface markdown-inline-code-face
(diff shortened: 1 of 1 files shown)


jrblevin:
Apologies for the delay in incorporating this, but I wanted to go back and check the commit to see what I was thinking.  Indeed, it was a typo. The idea was that all code faces (incl. inline code) would inherit from `markdown-code-face`. Thanks for catching this!

PR #257, opened 2017-09-18
Adds hook markdown-remove-gfm-checkbox-overlays

### LINKED ISSUE #238: Remove button overlays when exiting markdown-mode
Currently GFM checklists create buttons to toggle the value of the checkbox (which are really cool). The only problem is that these are overlays that persist when changing major mode, and can only be removed with `M-: (remove-overlays)`, which isn't user-friendly.

My suggestion is to add a hook to `after-change-major-mode-hook` to remove these overlays, which I think would be as simple as:
~~~elisp
(save-restriction
  (widen)
  (remove-overlays nil nil 'face 'markdown-gfm-checkbox-face))
~~~

I can submit a PR to fix this when I get the time.

rnkn:
Hook added to `change-major-mode-hook` when `markdown-make-gfm-checkboxes` is non-nil.

Closes #238

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -8594,6 +8594,13 @@ BEG and END are the limits of scanned region."
        (progn (goto-char beg) (beginning-of-line) (point))
        (progn (goto-char end) (forward-line 1) (point))))))
 
+(defun markdown-remove-gfm-checkbox-overlays ()
+  "Remove all GFM checkbox overlays in buffer."
+  (save-excursion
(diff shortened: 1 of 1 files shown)


jrblevin:
Many thanks!

rnkn:
Awesome, thanks for the merge. I had a look at how `org-mode` implemented removing overlays and it's more convoluted, but at heart still the same idea of a hook on `change-major-mode-hook`.

PR #262, opened 2017-09-23
Insert references before local variables

### LINKED ISSUE #216: C-c C-a r should place reference links above file-local variables
Create a new Markdown buffer with contents:

```markdown


<!-- Local Variables: -->
<!-- mode: markdown -->
<!-- End: -->
```

(the specific contents of the local variable section aren't important)

Then go to the beginning (M-<) and use C-c C-a r to insert a reference link. Link text and URL don't matter. The link will be inserted below the local variables section; it should be inserted above.

phst:
Fixes #216

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -4376,6 +4376,14 @@ be used to populate the title attribute when converted to XHTML."
       (immediately (markdown-end-of-text-block))
       (subtree     (markdown-end-of-subtree))
       (header      (markdown-end-of-defun)))
+    ;; Skip backwards over local variables.  This logic is similar to the one
+    ;; used in ‘hack-local-variables’.
+    (when (and enable-local-variables (eobp))
+      (search-backward "\n\f" (max (- (point) 3000) (point-min)) :move)
+      (when (let ((case-fold-search t))
+              (search-forward "Local Variables:" nil :move))
+        (beginning-of-line 0)
+        (when (eq (char-before) ?\n) (backward-char))))
     (unless (or (markdown-cur-line-blank-p)
                 (thing-at-point-looking-at markdown-regex-reference-definition))
       (insert "\n"))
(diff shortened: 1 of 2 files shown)


jrblevin:
Thanks for this (and your many other PRs)! Generally this seems fine, but I have two thoughts:

1. Shouldn't this only search for local variables in the range `(max (- (point-max) 3000) (point-min))` to `(point-max)`?  Ignoring the `(max ... (point-min))` part for simplicity, as I read the patch it looks from `(point)`, the proposed reference definition location to `(point)` minus 3000.  The difference is that I think it should start from `(point-max)` and probably only execute if the proposed point is somewhere in that range.

2. Since this is based on a simple string search, it would be "fooled" by a local variables block inside a code block, such as [this one](https://github.com/jrblevin/markdown-mode-guide/blob/650f2b6aa39743c69a05b1b2d88d531a630091b4/manuscript/guide.txt#L3214).  To protect against this, I think you could use `markdown-search-until-condition` with `markdown-conditional-search-function` let-bound to `search-backward` and with the condition being that the match is not inside a code block—something like`(lambda () (not (markdown-code-block-at-point-p)))`.

If the first one is fixed, that would at least minimize the chances of being affected by the second (i.e., only a code block near the end of the file would fool it).

phst:
> The difference is that I think it should start from (point-max) and probably only execute if the proposed point is somewhere in that range.

That should already be the case by virtue of checking for `eobp` in line 4381.

> Since this is based on a simple string search, it would be "fooled" by a local variables block inside a code block

True, but that's already the case for other uses of local variables. For example, try visiting a file containing

````markdown


``` html
<!-- Local Variables: -->
<!-- tab-width: 17 -->
<!-- End: -->
```
````

After visiting this file, `tab-width` will indeed be 17. So while not perfect, it's at least consistent with other uses of file-local variables.

jrblevin:
>> The difference is that I think it should start from (point-max) and probably only execute if the proposed point is somewhere in that range.
>
> That should already be the case by virtue of checking for eobp in line 4381.

You’re right, thanks. I missed that somehow.

PR #271, opened 2017-10-14
markdown-code-face is defined twice

alezost:
Hello, you created a duplicate of `markdown-code-face` by commit 7bf073664938f16a3bb7c0172db5102fd9874880, so I think the old one can be removed, right?


jrblevin:
Nice catch, thanks.

Issue #268, opened 2017-10-11
markdown-follow-link-at-point should be customizable for absolute filenames


phst:
In a link such as `[a](/b)`, `markdown-follow-link-at-point` will find the file `/b`. That's typically not the correct file because such paths are relative to some serving root. I guess their should either be a variable such as `markdown-link-file-root` or a function variable such as `markdown-get-file-for-url-function` to customize the behavior.

jrblevin:
Seems like a good idea to me. There is also the issue of file extensions, which comes up with wiki links as well. For wiki links, it is assumed that you want to open a source file with the same extension as the current file. I suppose the same would be preferred here, but it gets tricky when the URL might or might not have an extension. For example, either `[a](/b)` or `[a](/b.html)` might correspond to `b.md`. Given all the possibilities, maybe a function would be better here.

PR #277, opened 2017-10-29
Introduce user option to translate absolute filenames

phst:
## Related Issue

https://github.com/jrblevin/markdown-mode/issues/268

## Type of Change

## Checklist

- [x] I have updated the documentation at the beginning of `markdown-mode.el` if needed.
- [x] If documentation was added or changed, I have updated the **README.md** file using `webpage.sh`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,8 @@
     -   Save the buffer before running `markdown-open-command` and run
         `markdown-open-command` asynchronously.  Thanks to Dmitry
         Safronov for a patch.  ([GH-248][])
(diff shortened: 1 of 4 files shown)


jrblevin:
This looks fine to me, but before committing I wondered if it should be generalized to allow translating _all_ paths, not only absolute paths?  For example, one use case might be to add a file extension to an extensionless URL. Any thoughts?

phst:
Yes, that sounds simpler. Updated the PR.

jrblevin:
Great, thanks for the patch!

PR #281, opened 2017-11-05
Body wrapper customization. Fix #280

### LINKED ISSUE #280: Feature Request: Allow customization of body wrapper
Right now, the wrapper around the generated Markdown markup [isn't customizable](https://github.com/jrblevin/markdown-mode/blob/7a62259a22ccbf59b0132d077348ee489edf5d1e/markdown-mode.el#L7772), it would be nice if I could insert something after the open body tag and before the closing tag.

This would allow me to insert an article wrapper in order to apply a class name. This would allow me to use @sindresorhus's [github-markdown-css](https://github.com/sindresorhus/github-markdown-css) unmodified.

What do you think?

wyuenho:
This allows the addition of a body preamble and epilogue to further customize the generated HTML.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -7,6 +7,7 @@
 ;; Maintainer: Jason R. Blevins <jblevins@xbeta.org>
 ;; Created: May 24, 2007
 ;; Version: 2.4-dev
+;; Package-Version: 20171031.1725
 ;; Package-Requires: ((emacs "24") (cl-lib "0.5"))
 ;; Keywords: Markdown, GitHub Flavored Markdown, itex
 ;; URL: https://jblevins.org/projects/markdown-mode/
@@ -758,6 +759,14 @@
 ;;   * `markdown-xhtml-header-content' - additional content to include
 ;;     in the XHTML `<head>` block (default: `""`).
 ;;
+;;   * `markdown-xhtml-body-preamble' - additional content to include in the
+;;     XHTML <body> block, before the output. This is useful if you'd like to
+;;     wrap the Markdown output around extra elements (default: `""`')."
+;;
+;;   * `markdown-xhtml-body-epilogue' - additional content to include in the
+;;     XHTML <body> block, after the output. This is useful if you'd like to
+;;     wrap the Markdown output around extra elements (default: `""`').
+;;
 ;;   * `markdown-xhtml-standalone-regexp' - a regular expression which
 ;;     `markdown-mode' uses to determine whether the output of
 ;;     `markdown-command' is a standalone XHTML document or an XHTML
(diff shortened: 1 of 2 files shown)


markdown-mode.el
```diff
@@ -758,6 +759,14 @@
 ;;   * `markdown-xhtml-header-content' - additional content to include
 ;;     in the XHTML `<head>` block (default: `""`).
 ;;
+;;   * `markdown-xhtml-body-preamble' - additional content to include in the
+;;     XHTML <body> block, before the output. This is useful if you'd like to
+;;     wrap the Markdown output around extra elements (default: `""`')."
```
jrblevin: Stray single quote (`) here.

markdown-mode.el
```diff
@@ -758,6 +759,14 @@
...
+;;
+;;   * `markdown-xhtml-body-epilogue' - additional content to include in the
+;;     XHTML <body> block, after the output. This is useful if you'd like to
+;;     wrap the Markdown output around extra elements (default: `""`').
```
jrblevin: Stray single quote (`) here also.

markdown-mode.el
```diff
@@ -7,6 +7,7 @@
 ;; Maintainer: Jason R. Blevins <jblevins@xbeta.org>
 ;; Created: May 24, 2007
 ;; Version: 2.4-dev
+;; Package-Version: 20171031.1725
```
jrblevin: Please remove this `Package-Version` header.

markdown-mode.el
```diff
@@ -7770,7 +7791,13 @@ Standalone XHTML output is identified by an occurrence of
     (insert markdown-xhtml-header-content))
   (insert "\n</head>\n\n"
           "<body>\n\n")
+  (when (> (length markdown-xhtml-body-preamble) 0)
+    (insert markdown-xhtml-body-preamble))
+  (insert "\n")
```
jrblevin: It might be better to put this newline inside the `when` block, so that existing output for those who haven't customized this variable will be exactly the same.

markdown-mode.el
```diff
@@ -7770,7 +7791,13 @@ Standalone XHTML output is identified by an occurrence of
...
   (goto-char (point-max))
+  (insert "\n")
```
jrblevin: Same for this newline (see above).

jrblevin:
Here is a sketch of a test that I started working on. I didn't test the test yet though, but hopefully if it doesn't work as is it can be a good starting point.

``` lisp
(ert-deftest test-markdown-export/body-preamble ()
  "Test `markdown-xhtml-body-preamble'."
  (markdown-test-temp-file "export-test.md"
    (let* ((str "<!-- body preamble test -->")
           (markdown-xhtml-body-preamble str)
           (ofile (markdown-export))
           (obuffer (get-file-buffer ofile)))
      (with-current-buffer obuffer
        (goto-char (point-min))
        (should (search-forward str)))
      (kill-buffer obuffer)
      (delete-file ofile))))
```

wyuenho:
Ah I've completely forgotten about this PR. Thanks for fixing.

PR #285, opened 2017-11-14
Better gfm table recognition

Fuco1:
Add code to detect gfm tables.

## Description

in gfm tables do not require the leading `|` so they can simply look like this

    foo | bar
    ----|-----
    one | two

foo | bar
-----|-----
one | two

    asdasda | asdasd
    -|-
    onkjkjkjkje | two
    one | two
    -|-
    one | two

asdasda | asdasd
-|-
onkjkjkjkje | two
one | two
-|-
one | two

There is also a change of the table begin/end routines to use `markdown-table-at-point-p` instead of a special regexp to detect bounds. This makes the recognition depend only on one predicate instead of a regular expression which might vary between markdown implementations.

It is potentially slower than a regexp-based solution, but the performance hit is quite invisible in real-world usage---even on 1000 line tables it is pretty much instant on my 1.2ghz celeron CPU.

## Type of Change

## Checklist

- [ ] I have read the **CONTRIBUTING.md** document.
- [ ] I have updated the documentation at the beginning of `markdown-mode.el` if needed.
- [ ] If documentation was added or changed, I have updated the **README.md** file using `webpage.sh`.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -9142,6 +9142,12 @@ This version removes characters with invisibility property
 
 ;; Functions for maintaining tables
 
+(defvar markdown-table-at-point-p-function nil
+  "Function to decide if point is inside a table.
+
+The indirection serves to differentiate between standard markdown
+tables and gfm tables which are less strict about the markup.")
+
 (defconst markdown-table-line-regexp "^[ \t]*|"
   "Regexp matching any line inside a table.")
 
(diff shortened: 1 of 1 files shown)


jrblevin:
Thanks again!

Fuco1:
Hopefully this won't break anything :blush: I was trying to find some tests for tables but there were none.  Maybe I'll add another patch with some tests.

jrblevin:
No worries. I tested it informally and it worked fine for me. You're right that there aren't any table tests yet, but since it was a frequently requested feature I thought it was most beneficial to get the support included sooner and add tests later, and as any issues arise. If you write tests to include in the future, that would be great.

PR #301, opened 2017-12-30
Signal errors when executing ‘markdown-command’ or ‘markdown-open-command’ fails

phst:
Fixes #291

## Description

`markdown` and `markdown-open` now signal an error if the commands fail or the binaries don't exist.

## Related Issue

#291 

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Improvement (non-breaking change which improves an existing feature)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] I have read the **CONTRIBUTING.md** document.
- [x] I have updated the documentation in the **README.md** file if necessary.
- [x] I have added an entry to **CHANGES.md**.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed (using `make test`).

Issue #291, opened 2017-11-20
Suboptimal failure modes for markdown-open and markdown commands


phst:
## Expected Behavior

The commands `markdown-open` and `markdown` should provide a meaningful error message when `markdown-open-command` and `markdown-command` fail or don't exist, respectively.

## Actual Behavior

Both commands silently do nothing if the external subprocess fails. In addition, `markdown` silently does nothing if the program file doesn't exist.

## Steps to Reproduce

In a Markdown buffer:

```
(let ((markdown-command "/bin/false")) (markdown))
(let ((markdown-command "/does/not/exist")) (markdown))
(let ((markdown-open-command "/bin/false")) (markdown-open))
```

## Software Versions

- Markdown Mode: 2.3
- Emacs: 26.0.90
- OS: Debian testing

jrblevin:
Thanks, I agree!

PR #309, opened 2018-01-24
Fix "failed with exit code nil" when `markdown-open-command`

yewton:
## Description
* If DESTINATION is 0, `call-process' returns immediately with value nil
* We have to set DESTINATION nil and wait for PROGRAM to terminate so we can get exit code

## Related Issue

#291 #301

## Type of Change

## Checklist


jrblevin:
Good catch, thanks!

PR #320, opened 2018-03-15
skip empty URLs when inlining images

anarcat:
## Description

Markdown links like `[foo]()`, while basically invalid, do not usually
cause too much trouble, or at least shouldn't cause problems in
unrelated code. It so happens that if a document happens to have such
a link, `markdown-toggle-inline-images` will crash with:

    insert-file-contents-literally: not a regular file: /tmp/default-directory-path/

This patch fixes the problem by checking if the `file` path extracted
from the regex is empty before proceeding with treating it like an image.

## Type of Change

## Checklist

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -8376,7 +8376,8 @@ or \\[markdown-toggle-inline-images]."
         (let ((start (match-beginning 0))
               (end (match-end 0))
               (file (match-string-no-properties 6)))
-          (when (file-exists-p file)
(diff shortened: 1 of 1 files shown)


anarcat:
no new tests fail, but i haven't added test cases for this bug because there's no test case for `toggle-inline-images` already. there is already an empty link in `test-cases.txt` so once a test is added for the toggle, this will kick in automatically..

PR #343, opened 2018-06-07
Added 26.1 release

jjasghar:
Not sure if it's been pushed to travis yet, but 26.1 has been released.

Signed-off-by: JJ Asghar <jj@chef.io>


jrblevin:
It looks like 26.1 is indeed on Travis. Thanks for the patch!

PR #350, opened 2018-07-31
Change default character encoding

### LINKED ISSUE #340: HTML preview is not UTF-8
If I export the md file with 'markdown-preview', the specials characters are not correctly read on my browser (firefox 60.0.1)

## Expected Behavior
Export a md file containing letters with accents should be read in a browser supporting UTF-8.

## Actual Behavior
The html output should support UTF-8

## Steps to Reproduce
1. Create a markdown file containing letters with accents 
```markdown
This is special letter é
```
2. Export with 'markdown-preview' command
3. Open with firefox

## Software Versions
Markdown v2.1
Emacs 25.3.1

## Possible solution
I don't know much about HTML but I guess it misses a line before the title (markdown-mode.el:markdown-add-xhtml-header-and-footer :
```html
<meta charset=\"UTF-8\"/>
```
Also, I don't know why the html starts with an xml header specifing the UTF-8...

samihda:
I changed the default character encoding in the generated preview document to UTF-8.

No breaking changes were introduced, but I'm happy to adjust this in case we're okay with dropping `markdown-content-type` altogether and switching to the [`charset`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta) attribute instead.

## Related Issue

I came across this when text doesn't render properly in the preview. It was German text in my case, but I'm sure it affects others too. This should also fix #340.

## Type of Change

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -749,19 +749,19 @@ provides an interface to all of the possible customizations:
   * `markdown-css-paths` - CSS files to link to in XHTML output
     (default: `nil`).
 
-  * `markdown-content-type` - when set to a nonempty string, an
-    `http-equiv` attribute will be included in the XHTML `<head>`
-    block (default: `""`).  If needed, the suggested values are
-    `application/xhtml+xml` or `text/html`.  See also:
-    `markdown-coding-system`.
+  * `markdown-content-type` - used to set to the `http-equiv`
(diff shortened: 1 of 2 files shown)


jrblevin:
Thanks for the patch!

Issue #349, opened 2018-07-30
Infloop in jit-lock on emacs 27


vspinu:
When pasted into an empty markdown buffer the following sample cause an infloop:

------------------------

```
reg y x1 x2 x3, \\\
    vce(robust) \\\
    nocon
```
**Get**

```
reg y x1 x2 x3, \\\
vce(robust) \\\
nocon
```
----------------------

- Markdown Mode: ELPA/markdown-mode-20180707.1255
- Emacs: GNU Emacs 27.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30) of 2018-07-26
- OS: Ubuntu 18.04

First reported as part of vermiculus/magithub#353

jrblevin:
Thanks, but I wasn’t able to reproduce this. I copied the text between the horizontal rules when viewing the Markdown source of your comment. In case the spaces or newlines are important, perhaps you could email the test case to me as an attachment, upload it somewhere, or create a [Gist](https://gist.github.com/)?

jrblevin:
I just noticed you wrote Emacs 27, but I tested on Emacs 26.1. Does the version seem to be important?

vspinu:
Yes. It seems emacs 27. I cannot reproduce it on emacs 25. I will build emacs 26 and will check there as well.


And yet, copy pasting from the md source of my comment is what should trigger the error.

vspinu:
No, it's actually not emacs27 specific. i could reproduce it with emacs 25 as well, I just didn't notice the jit-lock error in the message buffer. 

I can reproduce it with the bare minimum emacs -q and:

```el
 (setq package-load-list '((markdown-mode t)))
 (package-initialize)
```

vspinu:
This is the minimal example I see it with:

-------
```
foo
```
**bar**
-----------------

jrblevin:
Thanks, now I also see the error in *Messages*.

vspinu:
I figured out the reason for this, in a nutshell it infolops because the position at the end of the fenced block is considered part of the code. I don't think this is right.  Let's say that I have an empty block `---\n---| ` where `|` stands for point. That's position 8 but the block has only 7 characters. 

Fine with changing this semantics?

vspinu:
Given that emacs buffer positions is forward-oriented (think of `get-text-property` and `put-text-property`) I think the only coherent semantics here is to consider the point after last fence outside of the code block.

[...]

PR #355, opened 2018-08-03
Don't consider point at the end of code-block as part of the code-block

vspinu:
## Description

A few reasons for the change.

Given that emacs is forward-oriented regarding text properties at point
(`get-text-property` retrieves text property from the forward char and
`set-text-property` is end-exclusive) this change makes quite a lot of sense.

From the user prospective, cursor is often identified with the point and given
that block-cursor is blinking on the next character it's confusing for the user
to treat that pistons as part of the preceding block.

At least one other tool (polymode) adopts similar conversion.

## Related Issue

 #349

## Type of Change

I guess users won't see a difference.

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,8 @@
         should fix the problem by following best practice and ensuring
         that your user configuration is loaded after the autoloads for
         `markdown-mode` are evaluated. ([GH-331][], [GH-335][])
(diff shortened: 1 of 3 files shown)


jrblevin:
Thanks very much for the patch! Apologies for not getting to it sooner.

PR #360, opened 2018-09-02
Remove polymode fix and fix boundary check in markdown-code-block-at-pos

vspinu:
Polymode was rewritten and it no longer interferes with markdown mode in any way. 

Regarding the boundary. In #355 I wrongly assumed that boundary can be either of the form `(123 456)` or nil. It turned out that that's not the case, it can be `(123 nil)` or even `(nil nil)`. Thus the current code in master throws invalid argument type on that comparison.

I am fixing this with a minimal check and passing the boundary downstream when second term is nil (like it used to be before), but I wonder if that's how this code should work. AFAICS the consumers of `markdown-get-enclosing-fenced-block-construct` and `markdown-code-block-at-pos` do seem to expect proper boundaries of the form `(123 456)`. So it's a bit surprising, in my mind, that things worked before :)

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -2870,12 +2870,7 @@ quoted code blocks.  Return nil otherwise."
         (let ((bounds (markdown-get-enclosing-fenced-block-construct pos)))
           (and bounds
                (< pos (cl-second bounds))
-               bounds))
-        ;; polymode removes text properties set by markdown-mode, so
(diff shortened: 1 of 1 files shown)


ybiquitous:
This patch works fine to me! 😄 🎉 
Please merge as soon as possible. 🙏

ybiquitous:
Thanks a lot! 😊

PR #373, opened 2018-11-07
Improve table insertion command

saf-dmitry:
## Description

- Fix column alignment specification
- Make specifying of columnt alignment more robust and user-friendly
- Refactor table insertion code

## Type of Change

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -579,13 +579,8 @@ can obtain a list of all keybindings by pressing <kbd>C-c C-h</kbd>.
     correctly when calculating column widths, however, columns
     containing hidden markup may not always be aligned properly.
 
-    <kbd>C-c C-s t</kbd> (`markdown-insert-table`) is a general command for
-    inserting new table. Specify table size, table align and table header in
-    minibuffer when executing `markdown-insert-table`.
-      - Specify row size.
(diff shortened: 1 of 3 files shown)


jrblevin:
There is an error (continuous integration checks failing) since the tests that were added in the previous implementation of this function are now failing, so I can't quite merge yet.

jrblevin:
Thanks for the PR and the update. The error was because the tests still weren't passing due to a small whitespace difference in the tables expected by the tests. The actual inserted tables were of this form:

```
|  |  |
|--:|--:|
|   |   |
|   |   |
```

But the tests were expecting them to be of this form:

```
|   |   |
|--:|--:|
|   |   |
|   |   |
```

I updated the code in d18a8f8 to add the extra space. I also updated the alignment prompt slightly and moved the point to the first column for text insertion.

Thanks again for the patch and let me know if you see any other issues.

saf-dmitry:
Now I see the problem: When a table is inserted at the first buffer line, then `markdown-table-begin` function has a problem to detect beginning of table (it needs at least one blank line before the table), and the first table row is not aligned properly. Maybe we should fix the `markdown-table-begin` function.

PR #404, opened 2019-06-30
Declare variable aliases before their referent

tarsius:
## Description

Since Emacs 27 the byte-compiler tells us to do that.
The warning is shown by `byte-compile-file-form-defvar-function',
which contains a comment providing this justification:

> Variable aliases are better declared before the corresponding variable,
> since it makes it more likely that only one of the two vars has a value
> before the `defvaralias' gets executed, which avoids the need to
> merge values.

## Type of Change

## Checklist

I do not think adding a test would make sense.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -2122,6 +2122,14 @@ Depending on your font, some reasonable choices are:
          ,@(when markdown-hide-markup
              `(display ,markdown-footnote-display))))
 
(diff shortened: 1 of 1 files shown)


tarsius:
Friendly ping.

(Note that this is a fairly trivial change that doesn't need much review.)

jrblevin:
Thank you for the patch!

PR #414, opened 2019-10-22
make preview more robust against renamed `eww` buffers

bymoz089:


syohex:
I suppose `eww-open-file` always creates `*eww*` buffer so we does not need to consider another buffer name.

PR #417, opened 2019-11-07
Autoload for mdwn files

sten0:
## Description

This PR Adds support for autoloading markdown-mode when a foo.mdwn file is visited.  It also includes a minor optimisation and cleanup.  Please let me know if you'd like me to rebase and drop any of these commits.

## Type of Change

## Checklist

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -9613,11 +9613,11 @@ rows and columns and the column alignment."
   (add-hook 'kill-buffer-hook #'markdown-live-preview-remove-on-kill t t))
 
 ;;;###autoload
-(add-to-list 'auto-mode-alist '("\\.markdown\\'" . markdown-mode))
-;;;###autoload
-(add-to-list 'auto-mode-alist '("\\.md\\'" . markdown-mode))
-;;;###autoload
-(add-to-list 'auto-mode-alist '("\\.mdwn\\'" . markdown-mode))
+(setq auto-mode-alist
+      (append auto-mode-alist
+              '(("\\.markdown\\'" . markdown-mode)
+                ("\\.md\\'" . markdown-mode)
+                ("\\.mdwn\\'" . markdown-mode))))
 
 
 ;;; GitHub Flavored Markdown Mode  ============================================
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -9616,6 +9616,8 @@ rows and columns and the column alignment."
 (add-to-list 'auto-mode-alist '("\\.markdown\\'" . markdown-mode))
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.md\\'" . markdown-mode))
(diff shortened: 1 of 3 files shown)


markdown-mode.el
```diff
@@ -30,7 +30,6 @@
 
 ;; See the README.md file for details.
 
-
```
anarcat: why did you remove the form feeds? it's actually quite common to have those in elisp modules. i never quite figured out *why*, that said, but i don't see how this relates to the rest of the (otherwise good) changes in this PR.

sten0: Ignorance and maybe OCD ;-)  I thought it was noise.  Since then I've read up on them.  Apparently they're sometimes (rarely) used as a kind of section marker.  I've rebased and dropped that commit.

sten0:
@jrblevin there's also a decent discussion on all the new extensions other applications are using for markdown here: https://bugs.freedesktop.org/show_bug.cgi?id=27441

I'm not sure which ones are the most popular ones are for Markdown 1.0-compatible files, but it might be worth documenting somewhere, in the case where adding a large chunk of them to the markdown-mode's autoload alist wouldn't be a good idea.

syohex:
I suppose it is better to use regexp instead of appending list as below

```lisp
(add-to-list 'auto-mode-alist '("\\.\\(?:markdown\\|md\\|mdwn\\)\\'" . markdown-mode))
```

markdown-mode.el
```diff
@@ -9613,9 +9613,11 @@ rows and columns and the column alignment."
   (add-hook 'kill-buffer-hook #'markdown-live-preview-remove-on-kill t t))
 
 ;;;###autoload
-(add-to-list 'auto-mode-alist '("\\.markdown\\'" . markdown-mode))
-;;;###autoload
-(add-to-list 'auto-mode-alist '("\\.md\\'" . markdown-mode))
+(setq auto-mode-alist
```
anarcat, line 9616: this looks strange - why do we reset the entire auto-mode-alist?

i prefered the previous way of doing multiple add-to-list (as in the readme) or using a regex

rnkn: Oh yes, very important *not* to set `auto-mode-alist` here, you'll break literally everyone's Emacses.

anarcat: well in this case we `setq` with the return value of `append` so it might just work. but yes, bad practice, i'd say.

rnkn: I see. Although this will add duplicates to `auto-mode-alist` whenever the file is evaluated. Defo better to use `add-to-list` (because it checks for duplicates first).

anarcat:
superseded by #443

syohex:
I have merged #443

sten0:
Cool, thank you!  Also, I agree the solution that was merged is better in every way :-)

PR #431, opened 2020-02-28
Avoid matching code fragments as code blocks

### LINKED ISSUE #403: gfm-mode does not treat ```no newline here``` correctly
## Expected Behavior

Given

    `one`
    ``two``
	```three```
	
`gfs-mode` should only set the `invisible` property to `markdown-markup` for all the backquotes but not for any of the other characters.

## Actual Behavior

The `invisible` property is also set to `markdown-markup` for the `three` part. (As expected that is *not* done for the `one` and `two` parts.)  At least for `gfm-mode`, strings that do not contain any newline characters and that are delimited with three backquotes on both sides should be treated the same as such strings are when they have only one or two backquotes on both sides.

I don't know whether using three backquotes that way is valid for official markdown, but it is for gfm.

Note that the `three` part above does not serve the same purpose as the `language` part below.

    ```language
	content
	```

The `language` part is markup; the `three` part is not.

## Steps to Reproduce

1. Copy the above first code-block above into a `gfm-mode` buffer.
2. Depending on how you did the copying, the three lines may begin with newline characters. If so, then remove those.
3. Evaluate `(add-to-invisibility-spec 'markdown-markup)` in that buffer.
4. Watch `three` become invisible but not `two` or `one`.

## Software Versions

Latest `markdown-mode` commit (115f77df9755c6a453f3e5d9623ff885d207ea82).

Emacs and OS versions are irrelevant.

## Other information

This is in reference to https://github.com/magit/forge/issues/45.

tarsius:
## Description

Code fragments should be delimited by one backquote on each side, but
some people use two or even three on each side.  Previously if three
were used then we wrongly identified that as the beginning of a code
block.  This resulted in the complete line being treated as syntax,
and hiding syntax therefore caused the complete line to be hidden.
Fix that by making sure no excess backquotes follow on the same line.

## Related Issue

Fixes #403.

## Type of Change

## Checklist

- [x] I have updated the documentation in the **README.md** file if necessary. (not necessary)

You are asking for quite a lot.  I could invest time figuring out how to write a test, but since I don't even know whether you see these pull-requests I skip doing that for now.  It would probably be much easier for you to write the tests, so I would very much appreciate it you could do that yourself. Considering that this is a one character change, I have already written an excessive amount of additional characters to justify why that character would be nice to have. ;-)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -109,6 +109,9 @@
 
 *   Bug fixes:
 
+    -   No longer treat code fragments that are delimited with three
+        backquotes on each side as the beginning of a code block.
+        ([GH-403][])
     -   Fix infloop caused by incorrect detection of end of code
         blocks ([GH-349][]).
     -   Remove GFM checkbox overlays when switching major modes.
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -720,7 +720,7 @@ Groups 1 and 3 match the opening and closing tags.
 Group 2 matches the key sequence.")
 
 (defconst markdown-regex-gfm-code-block-open
-  "^[[:blank:]]*\\(?1:```\\)\\(?2:[[:blank:]]*{?[[:blank:]]*\\)\\(?3:[^[:space:]]+?\\)?\\(?:[[:blank:]]+\\(?4:.+?\\)\\)?\\(?5:[[:blank:]]*}?[[:blank:]]*\\)$"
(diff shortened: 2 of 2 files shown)


tarsius:
I have just rebased this onto `master`, resolving the conflict.

syohex:
Please add test.

tarsius:
It will probably be a long time until I do that. It would be better and much faster if you did that. Sorry.

syohex:
I have applied your change at #492.

tarsius:
Thanks you!

Issue #319, opened 2018-03-09
Crash on the chinese Win10 with multimarkdown-6


alphatan:
When 'export', 'preview' a md file with markdown-mode with multimarkdown-6, it will consume all the memory then crash the OS (Windows)!

But running the multimarkdown command directly with the same file works fine.

## Expected Behavior

At least it should not crash the OS.

## Actual Behavior

The **multimarkdown** process consumes all the memory (16G) and crashes the OS (Win7, Win10).

## Steps to Reproduce

1. Running on Chinese Win7 or Win10
1. Installed multimarkdown-6 (6.3.0)
1. Load markdown-mode with following configurations
    ``` lisp
    (require 'markdown-mode)
    (setq markdown-command "multimarkdown") ;
    (autoload 'gfm-mode "markdown-mode"
       "Major mode for editing GitHub Flavored Markdown files" t)
    (setq markdown-fontify-code-blocks-natively nil)
    (setq markdown-xhtml-header-content "about 300K JS/CSS content")
    ```
1. Open any md doc (UTF-8 encoded)
1. Do '**export**' or '**preview**'
1. Then Cpu starts flying~~ (win7 will halt without any chance, while win10 will allow some while to kill the process)

## Backtrace

No backtrace could not be found  since the OS crashed.

## Software Versions

- Markdown Mode: From Git on 2017-10-17, 2.4-dev, and 2.3 (suffered a long time)
- Emacs: 25.3.1, 25.2, 225.1-2 (x86_64 from GNU)
- OS: Win7, Win10
- Multimarkdown: 6.3.0, 6.2.3, 6.2.2, 6.2.1, 5.4.0

## Temporary Solution

1. .emacs, setting encoding for multimarkdown process
    ``` lisp
    (when (eq system-type 'windows-nt)
      (set-default 'process-coding-system-alist
        '(
          ("[pP][lL][iI][nN][kK]" gbk )
          ("[mM][uU][lL][tT][iI][mM][aA][rR][kK][dD][oO][wW][nN]" utf-8 )
          ("[cC][mM][dD][pP][rR][oO][xX][yY]" gbk . gbk)
          )))
    ```
1. markdown-mode.el, change to call the multimarkdown.exe directly.
Find the line `(call-process-region`, then change to 
    ``` lisp
    (if (eq system-type 'windows-nt)
      (call-process-region begin-region end-region
                         markdown-command nil buf nil)
      (call-process-region begin-region end-region
[...]

PR #471, opened 2020-05-04
Use markdown-command directory instead of 'sh -c'

syohex:
## Related Issue

#319

## Type of Change

## Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -101,6 +101,7 @@
     -   If user set `display-buffer-alist` then use `display-buffer`. ([GH-413][])
     -   Add custom variable for opening image. ([GH-383][])
     -   Improve default `markdown-command` setting
+    -   Use markdown-command directly instead of 'sh -c' ([GH-319][])
 
 *   Bug fixes:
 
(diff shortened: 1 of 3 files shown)


jrblevin:
Hi @syohex, Thank you for so much hard work on `markdown-mode` lately. I have been strapped for time, so I really appreciate it. Users and contributors likely appreciate it even more!

I think quite a lot of users have arguments in a single string, as @Shiandow mentioned, myself included. It looks like this is backwards-compatible with arguments in a string, but also allows for them to be given in a list. Am I reading it correctly? In other words, both of these forms should work equally (including checking for the existence of the executable)?

```
(setq markdown-command "multimarkdown --snippet --smart --notes")
(setq markdown-command '("multimarkdown" "--snippet" "--smart" "--notes"))
```

syohex:
> In other words, both of these forms should work equally (including checking for the existence of the executable)?

Yes in that case. If user sets the argument which has spaces like

```lisp
(setq markdown-command "multimarkdown --some-option=\"This option has space\"")
```

then there do not work equally, however I suppose there are almost no people to use such configuration.

syohex:
I have added tests #482

PR #479, opened 2020-05-09
Fix markdown-command has command option case

syohex:
## Related Issue

https://github.com/jrblevin/markdown-mode/commit/67c2e8d9def80faad7e2041017526a7796e114b6

## Type of Change

## Checklist

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -7015,8 +7015,10 @@ The output buffer name defaults to `markdown-output-buffer-name'.
 Return the name of the output buffer used."
   (interactive)
   (save-window-excursion
-    (let ((begin-region)
-          (end-region))
+    (let* ((commands (and (stringp markdown-command) (split-string markdown-command)))
+           (command (car-safe commands))
+           (command-args (cdr-safe commands))
+           begin-region end-region)
       (if (use-region-p)
           (setq begin-region (region-beginning)
                 end-region (region-end))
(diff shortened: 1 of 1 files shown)


Shiandow:
Thanks, this fixes my problem for now.

Unfortunately this does to run into problems if the command or any of its arguments contains spaces.

Consider for example something like:
```lisp
  (setq markdown-command "pandoc.exe --from=markdown --to=html5 --metadata title=\"Title With Space\"")
```
and someone on windows is going to find out why you shouldn't have a path with spaces in it, but most windows users are probably used to that by now.

Honestly I'd be fine with it if markdown-command-args was a separate setting altogether, but this would break backwards compatibility so you might want to alert people to this if you encounter a command with spaces in it.

Or maybe you should reconsider what you were trying to solve and if this is the best way to do it.

I'm afraid I can't be of much help here, I've only just started using emacs (and elisp).

syohex:
I have merged #481 for supporting list of string. If you want to use such command and options, please set them as below.

```lisp
(setq markdown-command '("pandoc" "--from=markdown" "--to=html5" "--metadata=title:\"Title With Space\"") )
```

Shiandow:
Hi @syohex, 

Nice solution! Hope you didn't mind my nitpicking too much.

Like I said I only just started using emacs, so being able to diagnose this issue and not only build my own workaround but also get it fixed in the original source is a pretty amazing experience and is probably going to be one of the reasons I'll stick with it.

So thank you very much for your hard work.

PR #493, opened 2020-05-25
Provide filter-buffer-substring-function in gfm-view-mode

muffinmad:
## Description

`gfm-view-mode` is used by [eglot](https://github.com/joaotavora/eglot) to show symbol information in markdown format. E.g. the LSP server send something like this as the documentation:

<pre>
```python
foo(bar=None):
```

```
Docstring for foo
```
</pre>

`eglot-help-at-point` will show this text nicely fontified by `gfm-view-mode`.

The first line (which usually is the signature info) of that text is also used as the eldoc message to show in the echo area. The problem is that the first line is actually invisible codeblock marker.

This PR provides the function that return buffer substring without codeblock markers and can be used as value of the `filter-buffer-substring-function` variable. This way fontified signature info can be shown in the echo area.

Similar behavior is used by the `diff-mode` and the `backtrace-mode`.

Thanks!

## Type of Change

## Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -106,6 +106,8 @@
     -   Support list of strings of `markdown-command`
     -   Apply `markdown-translate-filename-function` for `markdown-display-inline-images`
         ([GH-422][])
+    -   Buffer substring filter function which filter out codeblock markers
+        is provided by the `gfm-view-mode`
 
 *   Bug fixes:
 
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -9517,11 +9517,22 @@ rows and columns and the column alignment."
   markdown-view-mode-map
   "Keymap for `gfm-view-mode'.")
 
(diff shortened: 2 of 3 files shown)


syohex:
@muffinmad Thanks for PR. However I suppose this is not enough

- `filter-buffer-substring-function` should be set both `markdown-view-mode` and `gfm-view-mode`
- It should filter not only code block(triple backticks) but also other invisible characters

I have sent PR #494. How about it ?

muffinmad:
> I have sent PR #494. How about it ?

Perfect! Thanks!

muffinmad:
Closing in favour of #494

syohex:
Thanks for checking. I have merged #494. Please check latest version.

muffinmad:
> Thanks for checking. I have merged #494. Please check latest version.

Works as expected. Great work! Thanks!

PR #540, opened 2020-09-02
Customizations for insert link

arthurcgusmao:
## Description

Two new features introduced to customize the behavior of
`markdown--insert-link-or-image`:

1. Option to define a `markdown-link-make-text-function` function that
(intelligently) defines a default link text before prompting the user.
2. Option to inhibit the prompt for a tooltip text via
`markdown-disable-tooltip-prompt`.

The reasoning for (1) is that someone can, for instance, have a function able to parse a webpage title, and then set `markdown-link-make-text-function` to that function to have an automated default link text in the prompt.

## Type of Change

## Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 *   New features:
     -   Introduce `markdown-mouse-follow-link` variable [GH-290][]
+    -   Option to define a `markdown-link-make-text-function` function
+        to automatically define a default link text before prompting the user.
+    -   Option to inhibit the prompt for a tooltip text via
+        `markdown-disable-tooltip-prompt`.
 
 *   Improvements:
     -   Cleanup test code
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -3715,6 +3715,28 @@ be used to populate the title attribute when converted to XHTML."
         "Reference [%s] was defined, press \\[markdown-do] to jump there")
        label))))
 
+(defcustom markdown-link-make-text-function nil
+  "Function that automatically generates a link text for a URL.
+
+If non-nil, this function will be called by
+`markdown--insert-link-or-image' and the result will be the
+default link text. The function should receive exactly one
+argument that corresponds to the link URL."
+  :group 'markdown
+  :type 'function
+  :package-version '(markdown-mode . "2.5"))
+
+(defcustom markdown-disable-tooltip-prompt nil
+  "Disable prompt for tooltip when inserting a link or image.
+
+If non-nil, `markdown-insert-link' and `markdown-insert-link'
+will not prompt the user to insert a tooltip text for the given
+link or image."
+  :group 'markdown
+  :type 'boolean
+  :safe 'booleanp
+  :package-version '(markdown-mode . "2.5"))
+
 (defun markdown--insert-link-or-image (image)
   "Interactively insert new or update an existing link or image.
 When IMAGE is non-nil, insert an image.  Otherwise, insert a link.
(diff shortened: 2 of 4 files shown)


arthurcgusmao:
For reference, here is an example of using the first improvement:

```lisp
(defun acg/url-get-page-title (url &optional timeout)
  "Returns the page title of an URL. If TIMEOUT (in seconds)
expires, return nil. Adapted from
https://lists.gnu.org/archive/html/help-gnu-emacs/2010-07/msg00299.html"
  (let ((timeout 1.2)
        (title))
    (with-current-buffer (url-retrieve-synchronously url t t timeout)
      (unless (= (buffer-size) 0)
        (goto-char (point-min))
        (re-search-forward "<title>\\([^<]*\\)</title>" nil t 1)
        (setq title (match-string 1))
        (goto-char (point-min))
        (re-search-forward "charset=\\([-0-9a-zA-Z]*\\)" nil t 1)
        (decode-coding-string title (intern (downcase
                                             (match-string 1))))))))

(setq markdown-link-make-text-function 'acg/url-get-page-title)
```

CHANGES.md
```diff
@@ -18,6 +18,16 @@
 
 # Markdown Mode 2.4
 
+*Sep 02, 2020*
+
+*   Improvements:
```
syohex, line 23: - Please add these change information to [Markdown Mode 2.5 section](https://github.com/jrblevin/markdown-mode/blob/master/CHANGES.md#markdown-mode-25)

markdown-mode.el
```diff
@@ -3711,6 +3711,21 @@ be used to populate the title attribute when converted to XHTML."
         "Reference [%s] was defined, press \\[markdown-do] to jump there")
        label))))
 
+(defvar markdown-link-make-text-function nil
```
syohex, line 3714: Please use `defcustom` for customizable variable instead of `defvar`

syohex (commented):
It looks your code base is a little old. Please update your source code or re-create PR with latest revision for avoid conflict.

arthurcgusmao:
Thank you for pointing that out @syohex (and also thanks for the specific comments). Indeed, I had not updated the package using `straight.el` before creating the spin-off. Hope things look good now.

CHANGES.md
```diff
@@ -36,6 +36,15 @@
   [gh-532]: https://github.com/jrblevin/markdown-mode/issues/532
   [gh-534]: https://github.com/jrblevin/markdown-mode/issues/534
 
+*Sep 02, 2020*
+
+*   Improvements:
+
+    -   Option to define a `markdown-link-make-text-function` function
```
syohex, line 43: Sorry I didn't explain it clearly. Add two lists at `New features` section. And the data(`*Sep 02, 2020*`) is not necessary, because it is release date and version 2.5 is not released yet.

arthurcgusmao: Ah I see, that makes sense. Changes should be grouped per 'stable release' version right? Please see if my latest commit fixes it.

syohex: I think your change introduces new features so I suppose they should be added at `New features` section as below image. How about ?

tmp

arthurcgusmao: My bad, I inserted the changes thinking they were in the new features section, sorry for the extra work.
Pushed a new commit now.

arthurcgusmao:
image
Now the checks are failing, but only the CHANGES.md file was modified so it should be related to some dependency that was unavailable when checking right? Is there a way to re-run the pipeline?

syohex:
I have re-run test, and it looks good.

syohex (approved):
Looks good.

@arthurcgusmao Would you like to clean up/squash commits before merging ? If so, I'll wait for it.

syohex:
Thanks. I have merged

PR #575, opened 2020-12-20
Support to display local inline images with spaces in file path encoded as %20

raphaelfeng:
Decode the file path with url-unhex-string before checking and loading the local file so that the images can be displayed.


syohex:
thanks

Issue #576, opened 2021-01-04
toggle local visibility before first heading (feature request?)


plantarum:
I often write markdown documents with a short preamble before the first heading:

```
---
title: "My Presentation"
---

Some preamble goes here.

# My first heading

Here's some text in the first section.
```

If I cycle visibility globally (`shift-TAB`), I get:

```
---
title: "My Presentation"
---
...
# My first heading...
```

This is good. I can open the first section by moving point to the heading line, and pressing `TAB`. However, I can't find a way to toggle visibility of the section prior to the first heading. I can display it again with global visibility cycling, but I wonder if there's a way to toggle local visibility cycling for this first section?

## Software Versions

- Markdown Mode: from melpa 2.5-dev
- Emacs: 28.0.50
- OS: Ubuntu 18

syohex:
It looks bug.

plantarum:
That's what I thought at first, but I'm not sure. If point is on a heading then I expect tab to cycle visibility. But in this case, there is no heading, so I'm not sure where point should be to enable the cycling? Or maybe the front section shouldn't be hidden ever? I don't know what the best approach is.

Before I found shift-tab for toggling global visibility cycling I was sure this must be a bug.

syohex:
@plantarum  I have sent PR #577. Please check it.

plantarum:
@syohex thanks, I'll take a look tomorrow!

PR #577, opened 2021-01-04
Fix markdown-shifttab for content where is after yaml header

syohex:
## Description

Don't fold content between YAML header end and first header

### Original version

before

### This patch version

after

## Related Issue

#576 (CC: @plantarum)

## Type of Change

## Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,6 +35,7 @@
     -   Fix wrong italic fontification just after code block [GH-548][]
     -   Fix too indended list face issue [GH-569][]
     -   Fix creating imenu index issue when there is no level-1 header too[GH-571][]
+    -   Fix markdown-shifttab for content after yaml header[GH-576][]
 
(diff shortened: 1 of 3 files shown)


plantarum:
This works, but doesn't address the question I raised in my issue. It might not have been clearly written, so that's not a surprise!

The problem as I originally encountered it was I would open a markdown file with a 'preamble' between the yaml block and the first heading. Based on my settings, all the headings were hidden, including the 'preamble'. I didn't know about shift-tab, and so couldn't find any way to unhide the top section.

Now that I know about shift-tab, I can see the preamble again.

The options we have now:

- original behaviour, with preamble folding only via global cycling (shift-tab)
- this pull, which means the preamble is **never** folded

What I'd really like is a way to toggle preamble folding locally. ie., to fold and unfold the preamble **without** folding and unfolding all other headings in the file. If that's not possible, I don't have a strong preference for the original behaviour or the new behaviour implemented here.

Thanks for your efforts!

PR #581, opened 2021-01-04
Support for braces in GFM fenced code blocks

plantarum:
Sorry, I hope you haven't been getting spammed with email every time I change this. I am not an expert git user.

## Summary

RMarkdown makes extensive use of language strings in the first line of GFM code blocks. ie.,

    ```{R, eval=TRUE, echo=FALSE}
    ls()
    ```

Markdown-mode correctly applies syntax highlighting to these blocks, but does not provide a function to insert blocks with braces. This pull request provides this feature as an option.

## Description

I have added the customization option `markdown-code-block-braces`, with a default value of `nil`. When this variable is set to `t`, code I have added to `markdown-insert-gfm-code-block` will insert the braces at the appropriate location. I also added a check to skip the insertion of space(s) after the fence in this case, as I think it looks funny to have a space at the beginning of the language string.

## Related Issue

If this looks appropriate for markdown, I would like to follow up with an additional convenience function to prompt the user for RMarkdown code block options and values. If that's too 'niche' for markdown-mode, I can look for another way to package it up.

## Type of Change

## Checklist

- [X ] I have read the **CONTRIBUTING.md** document.
- [ X] I have updated the documentation in the **README.md** file if necessary.
- [ X] I have added an entry to **CHANGES.md**.
- [ NA?] I have added tests to cover my changes.
- [ X] All new and existing tests passed (using `make test`).


syohex:
I have merged your changes into master branch.

PR #588, opened 2021-01-20
Closes #587 adds option to disable enumeration for ordered lists

### LINKED ISSUE #587: Add option to have ordered list use only '1. ' instead of enumerating
Almost all markdown viewers/renderers allow for the use of "1. " to denote a numbered list. so instead of 
``` 
1. item1
2. item2
3. item3
```
one can write
```
1. item1
1. item2
1. item3
```
which will then renders as:
``` markdown
1. item1
2. item2
3. item3
```

Examples that implement this behavior are sphinx / readthedocs, github, emacs markdown preview, etc. I strongly prefer this option since when people reshuffle their ordered lists, and they don't have a nice editor, they can't make mistakes.

I looked at the code of `markdown-insert-list-item` and on line `6040` is the relevant part:

``` elisp
            (let* ((old-prefix (match-string 1))
                   (old-spacing (match-string 2))
                   (new-prefix (if old-prefix
                                   (int-to-string (1+ (string-to-number old-prefix)))
                                 "1"))
```
changing the check just for `old-prefix` and adding a check for a configuration option that could be named:
`markdown-ordered-list-item-prefix-enumeration` (default `t`, and when `nil` doesn't enumerate, but always inserts "1. ") would be exactly the behavior I am looking for.

Would this be acceptable?

bklaase:
## Description
see: #587
tweaks `markdown-insert-list-item`
and `markdown-cleanup-list-numbers-level` to conform

## Related Issue
#587

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -427,6 +427,12 @@ The car is used for subscript, the cdr is used for superscripts."
   :group 'markdown
   :type 'string)
 
+(defcustom markdown-ordered-list-disable-enumeration nil
+  "Instead of enumerating ('1. 2. 3. etc.') only use '1. '.
+This will by most viewers renderers be accepted as a numbered list"
+  :group 'markdown
+  :type 'boolean)
+
 (defcustom markdown-nested-imenu-heading-index t
   "Use nested or flat imenu heading index.
 A nested index may provide more natural browsing from the menu,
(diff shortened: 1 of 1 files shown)


syohex:
Please do not remove the pull request template and please follow its rules. I have fixed this as #589. Could you check it ?

bklaase:
Thank you so much for going through the troubke. yes, I agree with any and all improvements and it works like this.

PR #592, opened 2021-02-02
Implement searching wiki link under project root

syohex:
## Description

https://github.com/jrblevin/markdown-mode/pull/591#issuecomment-768038622

## Type of Change

## Checklist


syohex:
All tests are passed with latest emacs on my local environment.

srid:
While this works with `project.el` ([as confirmed here](https://github.com/jrblevin/markdown-mode/pull/591#issuecomment-771831562)), I noticed that the experience is a bit jarring on doom-emacs. This is not markdown-mode's fault, because doom-emacs uses projectile and not project.el, which is perhaps why `C-c C-o` is repeatedly prompting me to specify the projecting root, rejecting what I give it most times, and then not remembering it for the rest of the session. I've asked in the Doom-Emacs discord channel asking how to best approach this problem; [here's the permalink](https://discord.com/channels/406534637242810369/406624667496087572/806281476525785158) to that question.

srid:
Just got a response, and it looks like this will solve the problem for Doom

> Imo the cleanest way would be to not use `(let ((roots (project-roots (project-current t)))) ... )`  but instead `(let ((roots (funcall markdown-project-roots (buffer-file-name))))` and have markdown-project-roots a defcustom variable that holds a function that returns a list of valid roots for a filename
> This way anyone can overwrite it properly (with project.el, projectile, or a function that is defined/advised in .dir-locals.el)

srid:
Actually, nevermind ... upgrading project.el resolves this issue. I only had to put this in doom-emacs's packages.el file.

```elisp
(package! project)
```

syohex:
I have committed a43629d. It let us not load project.el if project is managed by VCS system such as git.

And all tests are passed on all emacs version. Emacs 28 or higher versions requires `~/.emacs.d/projects` directory for using project.el functions. That directory does not exists on CI environment so tests were failed on Emacs 28.

PR #594, opened 2021-02-04
Silence byte-compiler

tarsius:
Starting with project v0.3.0 `project-roots` is obsolete and `project-root` should be used instead.

## Description

We change it so that `project-root` is used if available, else `project-roots`. Wrap a `with-no-warnings` around it to silence both new and old byte-compilers.

## Type of Change

## Checklist

- [x] *not necessary* I have updated the documentation in the **README.md** file if necessary.
- [ ] *that would be silly* I have added an entry to **CHANGES.md**.
- [ ] *that would be silly* I have added tests to cover my changes.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -7865,7 +7865,11 @@ The location of the alias component depends on the value of
                return it)
       (progn
         (require 'project)
-        (car (project-roots (project-current t))))))
+        (let ((project (project-current t)))
+          (with-no-warnings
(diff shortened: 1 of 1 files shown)


tarsius:
Jumping through the above hoops took long enough for me to forget to mention the following.

Instead of this you could also just use `project-root`, but then you should add `project-root` as an explicit dependency to ensure a new enough version is installed and you might not want to do that.

PR #606, opened 2021-03-22
Fix broken inline display of remote images

tejasvi:
## Description

#575 introduced a regression which prevents displaying the downloaded images (only `unhex_file` is checked, while the image is stored in `file`).

## Type of Change

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -8539,7 +8539,7 @@ or \\[markdown-toggle-inline-images]."
                     ;; strip query parameter
                     (setq file (replace-regexp-in-string "?.+\\'" "" file))
                     (unless (file-exists-p file)
-                      (setq file (file (url-unhex-string file))))))))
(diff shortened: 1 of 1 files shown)


syohex:
Could you tell us error case or add unit test for it ?

tejasvi:
I do not have enough experience to write tests (yet). The bug can be reproduced by inserting an image weblink (``) and then executing `markdown-display-inline-images` with `(setq markdown-display-remote-images t)`. In the code, notice that the `file` obtained from `markdown--get-remote-image` (L8538) is never used for rendering.

https://github.com/jrblevin/markdown-mode/blob/051734091aba17a54af96b81beebdbfc84c26459/markdown-mode.el#L8532-L8563

syohex:
Thanks for information. I'll check it later.

syohex:
@tejasvi sorry very belated reasponse. It looks good, thanks

PR #617, opened 2021-05-20
Fix delimiter regex in markdown-table-align

MrChenWithCapsule:
The regex in markdown-table-align to detect a delimiter row doesn't work when table delimiters are surrounded by spaces. This pull request fixes that issue and the expexted value of an existing test.

## Description

Consider the following table:
```md
| A |
| - |
|abcabcabcabc|
```

The original regex can not detect the delimiter row because the hyphen is surrounded by spaces.

## Type of Change

## Checklist

- [x] I have updated an existing test.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -9087,7 +9087,7 @@ This function assumes point is on a table."
             (indent (progn (looking-at "[ \t]*") (match-string 0)))
             ;; Split table in lines and save column format specifier
             (lines (mapcar (lambda (l)
-                             (if (string-match-p "\\`[ \t]*|[-:]" l)
+                             (if (string-match-p "\\`[ \t]*| *[-:]" l)
                                  (progn (setq fmtspec (or fmtspec l)) nil) l))
                            (markdown--split-string (buffer-substring begin end) "\n")))
             ;; Split lines in cells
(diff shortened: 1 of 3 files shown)


markdown-mode.el
```diff
@@ -9087,7 +9087,7 @@ This function assumes point is on a table."
             (indent (progn (looking-at "[ \t]*") (match-string 0)))
             ;; Split table in lines and save column format specifier
             (lines (mapcar (lambda (l)
-                             (if (string-match-p "\\`[ \t]*|[-:]" l)
+                             (if (string-match-p "\\`[ \t]*| *[-:]" l)
```
syohex, line 9090: Is tab not allowed ?

```suggestion
                             (if (string-match-p "\\`[ \t]*|[ \t]*[-:]" l)
```

tests/markdown-test.el
```diff
@@ -5921,7 +5921,7 @@ Details: https://github.com/jrblevin/markdown-mode/issues/308"
     (markdown-table-align)
     (should (string= (buffer-substring-no-properties (point-min) (point-max))
                      "| Col1 | Col2 |
-| :-:  | :-:  |
+|:----:|:----:|
```
syohex, line 5924: Could you add new tests for this fix instead of modifying existed one ?

markdown-mode.el
MrChenWithCapsule: I don't think tabs make much sense here, but allowing tabs does no harm.

tests/markdown-test.el
MrChenWithCapsule: Ok, I've added two tests, one for spaces and another for tabs.

syohex:
Thanks!!

PR #629, opened 2021-07-27
Sh shell

liuyinz:
## Description

When edit a code block with `sh-mode `(`bash`,`sh`,`zsh`,`ksh`,etc) in indirect buffer, the `sh-mode` always set default value of  `sh-shell` rather than according to `(markdown-code-block-lang)`

## Type of Change

## Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,7 @@
         name for `C-m`.
     -   Give `markdown-table-at-point-p-function` a non-nil default.
         Simplify `markdown-table-at-point-p` accordingly.
+    -   Auto set `sh-shell` when `markdown-edit-code-block` enter indirect buffer with `sh-mode`
 
 *   Bug fixes:
     -   Fix issue with `nil` being returned from `markdown-imenu-create-nested-index` [GH-578][]
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -47,8 +47,10 @@
 (defvar jit-lock-end)
 (defvar flyspell-generic-check-word-predicate)
 (defvar electric-pair-pairs)
+(defvar sh-ancestor-alist)
 
 (declare-function project-roots "project")
+(declare-function sh-set-shell "sh-script")
 
 
 ;;; Constants =================================================================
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -66,6 +66,7 @@
         instead of globally
     -   Modify only the buffer-local part of
         `syntax-propertize-extend-region-functions`.
+    -   Fix `sh-shell` error when `markdown-edit-code-block` enter indirec-buf with `sh-mode`
```
syohex, line 69:
```suggestion
    -   Fix `sh-shell` error when `markdown-edit-code-block` enter indirect-buf with `sh-mode`
```

markdown-mode.el
```diff
@@ -8784,6 +8785,11 @@ at the END of code blocks."
                       (lambda (_parent-buffer _beg _end)
                         (funcall mode)))
                      (indirect-buf (edit-indirect-region begin end 'display-buffer)))
+                (when (member mode '(shell-script-mode sh-mode))
+                  (with-current-buffer indirect-buf
+                    (if (member lang '("screen" "shell"))
+                        (sh-set-shell "sh")
+                      (sh-set-shell lang)))) ;; reset `sh-shell' in indirect buffer
```
syohex, line 8792: Should `lang` be checked as valid shell ? Some code blocks uses `shell-script`. This is not valid shell name. And some shell like `zsh` or `fish` does not recognize on Windows.

https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

CHANGES.md
syohex, line 69: I'm not sure that is this bug ? I suppose this is improvement.

markdown-mode.el
liuyinz: I use `zsh` as $SHELL on MacOS, however , when I edit `sh` or `bash` code block, the indirect-buffer always identified as `zsh`. Of course , I could set `sh-shell-file` to `bash` as default , but when I edit `zsh` code block, It will identified as `bash` .

liuyinz:
I update the conditions, have a look.

- Non-MS-Windows system
- belong to `sh-mode`, which is classified by `markdown-code-lang-modes`
- valid shell type check

otherwise, it will still use `sh-shell-file`'s default value

syohex (approved):
Looks good. Could you squash commits ?

syohex:
@liuyinz Thanks for your work.

Issue #638, opened 2021-08-12
`markdown-fill-paragraph` needs to respect setext headings


Knusper:
## Expected Behavior

For this

```
Heading
=======
```

When the cursor is somehwere in there `markdown-fill-paragraph` should do nothing.

## Actual Behavior

But after `markdown-fill-paragraph` we have

```
Heading =======
```

- Markdown Mode: 2.5-dev
- Emacs: 26.1
- OS: debian 10

syohex:
I have merge #645. Could you try to latest version ?

Knusper:
This works.

PR #645, opened 2021-08-14
Fix fill paragraph with setext heading issue

syohex:
## Related Issue

#638

## Type of Change

## Checklist

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -9730,6 +9730,7 @@ rows and columns and the column alignment."
                            ;; not paragraph-ending suffixes:
                            ".*  $" ; line ending in two spaces
                            "^#+"
+                           "^[-=]+"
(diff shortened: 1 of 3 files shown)


markdown-mode.el
```diff
@@ -9730,6 +9730,7 @@ rows and columns and the column alignment."
                            ;; not paragraph-ending suffixes:
                            ".*  $" ; line ending in two spaces
                            "^#+"
+                           "^[-=]+"
```
gbordyugov, line 9733: This seems to be breaking paragraph fill of list elements started by a `-` at the beginning of the line, is this behaviour intended?

syohex: Thanks for comment. I didn't consider about list syntax. I'll fix it later.

gbordyugov: Thanks a lot for answering so quickly!

PR #674, opened 2022-01-10
fix bug that fails to display link in eldoc when url-hidding mode is enabled

taquangtrung:
Hi,

## Description

Currently, when the `URL hiding` feature is enabled, `markdown eldoc` always throws the error `eldoc error: (wrong-type-argument stringp nil)` when a cursor is put under the hidden URL.

I investigated the code and see that the problem is due to the fact that the regular expression matching state performed by `(thing-at-point-looking-at markdown-regex-link-inline)`, which is used to get the URL, is cleared by the next call to `(markdown--substitute-command-keys (if imagep "\\[markdown-insert-image]" "\\[markdown-insert-link]"))`.

So, this PR will fix this bug by getting the URL information early, before the call to `markdown--substitute-command-keys`.

Can anyone take a look and merge it if possible?

Thank you!

## Type of Change

## Checklist

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -9645,22 +9645,21 @@ rows and columns and the column alignment."
              (thing-at-point-looking-at markdown-regex-link-reference))
          (or markdown-hide-urls markdown-hide-markup))
     (let* ((imagep (string-equal (match-string 1) "!"))
+           (referencep (string-equal (match-string 5) "["))
+           (link (match-string-no-properties 6))
            (edit-keys (markdown--substitute-command-keys
                        (if imagep
                            "\\[markdown-insert-image]"
                          "\\[markdown-insert-link]")))
            (edit-str (propertize edit-keys 'face 'font-lock-constant-face))
-           (referencep (string-equal (match-string 5) "["))
            (object (if referencep "reference" "URL")))
       (format "Hidden %s (%s to edit): %s" object edit-str
               (if referencep
                   (concat
                    (propertize "[" 'face 'markdown-markup-face)
-                   (propertize (match-string-no-properties 6)
-                               'face 'markdown-reference-face)
+                   (propertize link 'face 'markdown-reference-face)
                    (propertize "]" 'face 'markdown-markup-face))
-                (propertize (match-string-no-properties 6)
-                            'face 'markdown-url-face)))))
+                (propertize link 'face 'markdown-url-face)))))
    ;; Hidden language name for fenced code blocks
    ((and (markdown-code-block-at-point-p)
          (not (get-text-property (point) 'markdown-pre))


syohex:
Could you show us the code which reproduces your issue or add unit tests ?

taquangtrung:
The code that creates the issue is inside `markdown-eldoc-function`, in particular this part:

``` emacs-lisp
   ((and (or (thing-at-point-looking-at markdown-regex-link-inline)
             (thing-at-point-looking-at markdown-regex-link-reference))
         (or markdown-hide-urls markdown-hide-markup))
    (let* ((imagep (string-equal (match-string 1) "!"))
           (edit-keys (markdown--substitute-command-keys
                       (if imagep
                           "\\[markdown-insert-image]"
                         "\\[markdown-insert-link]")))
           (edit-str (propertize edit-keys 'face 'font-lock-constant-face))
           (referencep (string-equal (match-string 5) "["))
           (object (if referencep "reference" "URL")))
      (format "Hidden %s (%s to edit): %s" object edit-str
              (if referencep
                  (concat
                   (propertize "[" 'face 'markdown-markup-face)
                   (propertize (match-string-no-properties 6)
                               'face 'markdown-reference-face)
                   (propertize "]" 'face 'markdown-markup-face))
                (propertize (match-string-no-properties 6)
                            'face 'markdown-url-face)))))
```

The function call `(thing-at-point-looking-at markdown-regex-link-inline)` is supposed to be used to fetch URL information later by `(match-string-no-properties 6)`.

But there is another call by `markdown--substitute-command-keys` at 

``` emacs-lisp
 (edit-keys (markdown--substitute-command-keys
                       (if imagep
                           "\\[markdown-insert-image]"
                         "\\[markdown-insert-link]")))
```

which also performs regular expression matching. This call will clear the state produced by `(thing-at-point-looking-at markdown-regex-link-inline)`.

Therefore the latter call to `(match-string-no-properties 6)` always produces `nil`.

In my patch, I put the call to `(match-string-no-properties 6)` to be earlier than `markdown--substitute-command-keys`

syohex:
Sorry I would like you to show us sample markdown text and minimal configuration to reproduce this issue. And could you tell us how  to reproduce in detail, cursor position, commands etc ?

taquangtrung:
Ah I see. For example, I created and open a markdown file `foo.md` and put the following content into it: `[This is a hidden link](http://hidden.link)`.

Then run `M-x markdown-toggle-url-hiding`, turn on `eldoc-mode`, and put the cursor under the link's text. There will be an error message like the below:

markdown

taquangtrung:
Here is what obtained after applying my patch:

markdown-new

syohex:
@taquangtrung Thanks for good explanation. I have merged your patch and added a unit test #675

taquangtrung:
@syohex: thanks for the update!

PR #711, opened 2022-06-20
Fix functions with multiple docstrings

tarsius:
## Description

A function should have a single docstring.  There are two functions that have two docstrings each, for each of these functions I remove one of the docstring.  In both cases the bad docstring was in the correct place, so I don't only remove the bad docstring, I also move the good docstring to the correct location.

(Actually the "good docstrings" aren't that great either.  The first line of a docstring should end with a period.  I have not fixed this.)

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] *Not necessary* I have updated the documentation in the **README.md** file if necessary.
- [X] *I.e. not necessary* I have added an entry to **CHANGES.md**.
- [X] *I.e. not necessary* I have added tests to cover my changes.

## PS

Do I have your permission to forgo filling out this questionnaire for equally trivial fixes in the future?

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -7678,20 +7678,18 @@ displaying the rendered output."
          (setq markdown-live-preview-source-buffer nil))))
 
 (defun markdown-live-preview-switch-to-output ()
-  "Switch to output buffer."
-  (interactive)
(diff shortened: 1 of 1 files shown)


syohex:
thanks

Issue #578, opened 2021-01-04
"Wrong type argument: consp, nil" from imenu


sandinmyjoints:
## Expected Behavior
- imenu works

## Actual Behavior
- `imenu--make-index-alist: Wrong type argument: consp, nil`

## Steps to Reproduce

1. call `imenu` on a markdown-mode buffer with the following content:

```gfm
# A

### B

## C
```

## Software Versions

- Markdown Mode: 2.5-dev
- Emacs: 28.0.50
- OS: macOS 10.14.6

syohex:
I cannot reproduce this issue with both Emacs 28.0.50 and 27.1. 

I did

- `emacs -Q -l markdown-mode.el test.md`
- `M-x imenu`

Then no error happens. Could you tell us how to reproduce this issue in more detail ? cursor position, minimum configuration etc.

rien333:
Are you by any change using the `native-comp` branch from emacs 28? I had a similar issue with that branch (I don't recall how I fixed it at the moment, however)

sandinmyjoints:
I am, yes! Would definitely be interested in your fix.

rien333:
I think that the following were among the steps that restored `imenu` to an usable state for me (in the context of `markdown-mode`):

1. Make sure to have an up-to-date emacs 28
2. Add `(setq markdown-nested-imenu-heading-index nil)` to your init.el

Why this works, I don't know. Could be that I did more stuff to fix this.

wyuenho:
Does this issue really have to do with emacs 28?

rien333:
Yes, but only in so far emacs 28 introduced the native comp feature, which at one point broke some imenu stuff.

syohex:
I still cannot reproduce this issue with emacs master branch(2021/04/30) with native comp build flag.

rien333:
@syohex could be that this (or some underlying) issue has been fixed by now, especially since the native-comp branch is under heavy development.  In any case, I can't reproduce this issue anymore either.

syohex:
Thanks for information. Please reopen if this issue happens again

wyuenho:
I found this issue before I ran into it, please reopen.

liuyinz:
yes, several days ago,the problem seems fixed by emacs28,but it appeared again after I update the native-comp branch...

syohex:
@wyuenho please try to latest code. And this issue is not markdown-mode issue. If you still have problem, please report issue to emacs development mailing list.

[...]

PR #721, opened 2022-07-22
Fixes bug related to interaction with emacs's native compilation

psibi:
## Description

This change specifically fixes this bug for me:
https://github.com/ardumont/markdown-toc/issues/51

IIUC, `#` is used to inform the byte-compiler that the expression can
be byte-compiled. But in this case there is nothing to compile since
it's a symbol.

## Related Issue

https://github.com/ardumont/markdown-toc/issues/51

## Type of Change

## Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,10 +18,13 @@
 
 *   Bug fixes:
     - Don't override table faces by link faces [GH-716][]
+    - Don't bytecompile symbol unless it's required for imenu based
+      functions. This fixes [mt-51][]
 
   [gh-572]: https://github.com/jrblevin/markdown-mode/issues/572
   [gh-705]: https://github.com/jrblevin/markdown-mode/issues/705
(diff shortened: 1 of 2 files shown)


psibi:
This doesn't seem to be the right fix. The bug has returned back. If I restart emacs, it doesn't happen again. This needs more fiddling.

syohex:
I suppose this is same as #578. Workaround is https://github.com/jrblevin/markdown-mode/issues/578#issuecomment-1126380098

psibi:
@syohex Thanks, I didn't know about that. Is there a more proper fix for it instead of opting out of native compilation for markdown mode ?

syohex:
>  Is there a more proper fix for it instead of opting out of native compilation for markdown mode ?

I have no idea yet.

psibi:
Cool, thanks for the update. The workaround would keep me sane. :-)

PR #759, opened 2023-04-08
Fix byte-compile warnings

syohex:
## Type of Change

## Checklist

- [ ] I have read the **CONTRIBUTING.md** document.

### DIFF
--- a/tests/markdown-test.el
+++ b/tests/markdown-test.el
@@ -57,7 +57,8 @@
            (setq-default indent-tabs-mode nil)
            (goto-char (point-min))
            (font-lock-ensure)
-           (prog1 ,@body (kill-buffer))))))
+           (prog1 ,@body (kill-buffer)))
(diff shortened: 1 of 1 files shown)


mattiase:
While this makes the warning go away (for now), you should probably think about what that warning was trying to tell you.
`unwind-protect` is completely pointless without something meaningful in the unwinding part. If the `unwind-protect` isn't needed, as in this case, you can just remove it.

You may also want to take a look at the `(prog1 ,@body (kill-buffer))` part. The reason for using `prog1` is to return the value of the first argument, but which one is that? The `,@body` is a splice-in.
Fortunately, the call to `kill-buffer` can be removed as well because `with-temp-buffer` will do that, so the `prog1` is no longer needed. You can also remove the call to `erase-buffer` because `with-temp-buffer` always gives you an empty buffer.

PR #793, opened 2023-08-16
Push mark before jumping to footnote

slotThe:
## Description

When inserting a (small) footnote, it is often very convenient to jump back to the main text with C-u C-SPC instead of using `markdown-footnote-return`. However, this can only be done if `markdown-insert-footnote` pushes to the mark ring before jumping, so do that.

## Type of Change

- [n/a] Bug fix (non-breaking change which fixes an issue)
- [n/a] New feature (non-breaking change which adds functionality)
- [n/a] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [n/a] I have updated the documentation in the **README.md** file if necessary.
- [n/a] I have added an entry to **CHANGES.md**.

  Not significant enough to warrant this, I think.

- [n/a] I have added tests to cover my changes.

  Not significant enough to warrant this, I think.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -4665,6 +4665,7 @@ element. More details here https://developer.mozilla.org/en-US/docs/Web/HTML/Ele
   (interactive)
   (let ((fn (markdown-footnote-counter-inc)))
     (insert (format "[^%d]" fn))
+    (push-mark (point) t)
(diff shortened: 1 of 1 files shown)


syohex:
`org-footnote-new` sets marker like this so I think this behavior looks good.

PR #812, opened 2023-11-10
Add option markdown-need-active-region

snan:
This has been driving me crazy since I got used to working on just
normal regions the old-fashioned way.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -154,6 +154,11 @@ promotion and demotion functions."
   :group 'markdown
   :type '(repeat string))
 
+(defcustom markdown-need-active-region t
+  "Require region to be active for markup insertion commands."
+  :group 'markdown
+  :type 'boolean)
+
 (defcustom markdown-bold-underscore nil
   "Use two underscores when inserting bold text instead of two asterisks."
   :group 'markdown
@@ -3760,7 +3765,7 @@ prefixed with an integer from 1 to the length of
 
 (defun markdown--insert-common (start-delim end-delim regex start-group end-group face
                                             &optional skip-space)
-  (if (use-region-p)
+  (if (or (use-region-p) (not markdown-need-active-region))
       ;; Active region
       (let* ((bounds (markdown-unwrap-things-in-region
                       (region-beginning) (region-end)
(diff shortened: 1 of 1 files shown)


syohex:
I understand your opinion. However I think this is not a natural emacs way and no one uses this option except you. Please override `use-region-p` or `region-active-p` by yourself as below. You could control behaviors other than markdown-mode by this way.

```
(defun use-region-p ()
  (and (or (bound-and-true-p markdown-mode) (region-active-p)) ;; customize as you like
       (or (> (region-end) (region-beginning))
           (and use-empty-active-region
                (not (eq (car-safe last-input-event) 'down-mouse-1))
		(not (mouse-movement-p last-input-event))))))

;; or
(defun region-active-p ()
  (and mark-active
       (progn (cl-assert (mark)) t)))
```

snan:
Setting mark with `C-SPC`, moving the pointer to operate the region, and then doing something in the region is not a natural Emacs way?! But that's how every other mode works. `comment-region`, `rot13-region`, `eval-region`, `shell-command-on-region`, the only mode that does *not* work this way is markdown-mode and it used to work that way in the past.

snan:
The reason markdown-mode got its new unusual behavior was so that the same commands would work for word-at-point and for the region itself. That's the benefit of doing it the "active regions only" way. For me personally that benefit is cool but definitively not worth it.

That's why I made it a setting. I don't even know how you are using it as is, do you all have transient mark on or what?

snan:
Some packages like [vertico](https://github.com/minad/vertico) have this `(or (use-region-p) (not transient-mark-mode))`, that's maybe a better idea than a new option.

trentbuck:
A bit of context:

AIUI `use-region-p` works well iff the user has `transient-mark-mode` enabled.
It was disabled by default until GNU Emacs 23 (about 16 years ago).
Having `(transient-mark-mode -1)` is not so much "unnatural" as "traditional" -- or perhaps "legacy" ;-)
I don't know how reasonable it is to put `(defun use-region-p nil t)` next to `(transient-mark-mode -1)`.

https://www.gnu.org/software/emacs/manual/html_node/emacs/Disabled-Transient-Mark.html
https://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS.23#n1013

PR #813, opened 2023-11-13
Introduce ignore region customize variable

syohex:
## Description

The variable name is same as org-mode's one

## Related Issue

#812 (CC: @snan )

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves an existing feature)
- [s] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] I have read the **CONTRIBUTING.md** document.
- [x] I have updated the documentation in the **README.md** file if necessary.
- [x] I have added an entry to **CHANGES.md**.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed (using `make test`).

PR #814, opened 2023-11-14
Check if transient-mark-mode is on

snan:
This doesn't make sense for all commands like header insertion, so
I only implemented it for some commands. Some of them said in the
documentation that they check if transient-mark-mode was on which
they didn't seem to do but they do with this commit. This is the way
vertico and a couple of other packages use (use-region-p) in a more
legacy-friendly way.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -3665,7 +3665,7 @@ and S1 and S2 were only inserted."
             b end
             new-point (+ (point) (length s1))))
      ;; Active region
-     ((use-region-p)
+     ((or (use-region-p) (not transient-mark-mode))
       (setq a (region-beginning)
             b (region-end)
             new-point (+ (point) (length s1))))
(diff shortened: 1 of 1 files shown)


snan:
This replaces #812 and #813 without introducing a new option

syohex:
This is not correct because the region might not be defined when `transient-mark-mode` is nil. Then region functions such as `region-beginning`, `region-end` throw an error.

I have read lots of elisp code but they use `use-region-p` and don't care `not transient-mark-mode` case. You gave `vertico` as an example, but it doesn't use region functions around checking `not transient-mark-mode`.

I suppose you should do one of the followings
- override `use-region-p` as you like
- use `Ctrl-SPC Ctrl-SPC` (enable transient-mark-mode once) when you want to use some markdown command with the region
- Enable `transient-mark-mode` and disable its face attributes

Or please fork this repository and improve it as you like, if you seriously need this feature,

PR #826, opened 2024-03-09
Hide trailing whitespace when hiding markup

jimporter:
## Description

When hiding markup, e.g. via `markdown-view-mode`, the trailing whitespace markup for line breaks doesn't get hidden. This patch changes it so that the whitespace *is* hidden when hiding other markup. In practice, this just makes the `"  "` invisible and nothing else; then, we let the newline do its usual thing.

(In the future, I could imagine `markdown-view-mode` handling soft newlines differently, so that something like `foo\nbar` is displayed as `foo bar`. Then, we'd probably want to do something special to mark the newline in `  \n` as a *hard* newline.)

## Related Issue

See <https://debbugs.gnu.org/cgi/bugreport.cgi?bug=69647>

## Type of Change

## Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@
     - Add functions to move to the beginning and end of lines
       (`markdown-beginning-of-line` and `markdown-end-of-line`), and the
       variable `markdown-special-ctrl-a/e`, like Org mode.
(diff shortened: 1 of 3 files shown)


syohex:
Thanks

Issue #799, opened 2023-09-29
Markdown mode leads to heavy input lag


RalfJung:
Since a recent emacs update, markdown mode leads to heavy input lag when editing inside a list. This makes it basically unusable, I've uninstalled it for now.

## Expected Behavior

Everything I type should appear on the screen immediately.

## Actual Behavior

There is noticeable input lag, making editing quite painful.

## Steps to Reproduce

- open a markdown file like https://gitlab.mpi-sws.org/iris/stdpp/-/blob/master/CHANGELOG.md?ref_type=heads
- go to somewhere inside the first list of changes, and start typing

## Software Versions

- Markdown mode: 20230830.1338
- emacs: 29.1+1-5
- OS: Debian testing

tenpast:
I can confirm this issue. Running emacs profiler I get the following results:

- Memory-Profiler-Report

````
     42,604,780  81% + redisplay_internal (C function)
      8,102,036  15% + command-execute
      1,833,645   3% + jit-lock--antiblink-post-command
         43,728   0% + timer-event-handler
          4,144   0%   internal-echo-keystrokes-prefix
             24   0% + eldoc-schedule-timer
              0   0%   ...
````

- CPU-Profiler-Report

````
       24573  94% + redisplay_internal (C function)
        1122   4% + command-execute
         115   0% + ...
          37   0% + jit-lock--antiblink-post-command
          22   0% + timer-event-handler
````

**Comment by umanwizard** [comment]
I can reproduce in any buffer with a lot of `_` characters enclosed in backticks. For example, generate a buffer as follows:

```
(dotimes (n 1000)
  (insert "`my_test_string`\n"))
```

then observe that running `(font-lock-debug-fontify)` takes a long time (12 seconds for me, and also appears to be quadratic in the number of lines).

The heavy time-takers seem to be `markdown-match-italic` and, in particular, its call to `markdown-inline-code-at-pos-p`.

umanwizard:
For some reason I can't explain, the loss-of-interactivity issue can be worked around by turning on `jit-lock-debug-mode` in the buffer.

jdtsmith:
[...]

Issue #391, opened 2019-04-25
Performance issues in long lists when using italics.


Mihara:
Linux, Emacs 26.2, current markdown-mode MELPA release.

When editing a list of 30+ items, each of them a few lines long with at least one word in italics, performance degrades spectacularly in linear proportion to the number of items in the list, eventually resulting in latencies approaching 0.5s between a keypress and it showing up on screen.

Let us have a file like that:

```
+ So I have a markdown file, and inside I have a long-ish list of items, each a paragraph long, or long enough to wrap around, using some *italics*.
```

Repeat this line 30-50 times. The slowdown increases proportionally to the number of items in the list, and it persists with `emacs -Q`, though obviously, it takes a few more lines to achieve the same kind of slowdown. (76 is sufficient in my case.)

`profiler-start` reports something like this:
```
- flyspell-post-command-hook                                     2347  52%
 - if                                                            2347  52%
  - progn                                                        2347  52%
   - condition-case                                              2336  52%
    - let                                                        2336  52%
     - let                                                       2336  52%
      - if                                                       2336  52%
       - flyspell-check-word-p                                   2332  52%
        - let                                                    2332  52%
         - cond                                                  2332  52%
          - cond                                                 2332  52%
           - and                                                 2332  52%
            - sit-for                                            2332  52%
             - redisplay                                         1893  42%
              - redisplay_internal (C function)                  1873  42%
               - jit-lock-function                               1857  41%
                - jit-lock-fontify-now                           1857  41%
                 - jit-lock--run-functions                       1857  41%
                  - run-hook-wrapped                             1857  41%
                   - #<compiled 0x284ae81>                       1857  41%
                    - font-lock-fontify-region                   1853  41%
                     - font-lock-default-fontify-region               1853  41%
                      - font-lock-fontify-keywords-region               1837  41%
                       - markdown-match-italic                   1701  38%
                        - markdown-inline-code-at-pos-p               1669  37%
                         - markdown-inline-code-at-pos               1665  37%
                          - markdown-match-code                  1426  32%
                           - markdown-search-until-condition               1426  32%
                              apply                              1426  32%
                            markdown-beginning-of-text-block                199   4%
                            markdown-end-of-text-block                 36   0%
                        + markdown-match-inline-generic                 32   0%
                       + markdown-fontify-plain-uris                 28   0%
```

[...]

PR #842, opened 2024-08-19
Get rid of some quadratic behavior

umanwizard:
markdown-match-bold and markdown-match-italic both call markdown-inline-code-at-pos-p on each candidate match, which takes linear time in the distance
from the start of the block to the point being checked.

In a large block with many candidate matches inside inline blocks, this is slow because the function is called with each candidate, resulting in overall quadratic behavior, which this commit fixes by starting the code-at-pos scan after the last already-found match.

For example, the following benchmark code:

```
(with-temp-buffer
  (dotimes (_ 400)
    (insert "`my_test_string`\n"))
  (markdown-mode)
  (car (benchmark-run
           (font-lock-debug-fontify))))
```

takes about 10.5 seconds before this commit, and about 0.14 seconds after this commit.

## Type of Change

## Checklist

(except `test-markdown-export/buffer-local-css-path` which was already failing)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,6 +63,8 @@
     - Consider `major-mode-remap-alist` to determine major-mode for code blocks [GH-787][]
     - Set marker after footnote reference [GH-793][]
     - Improve putting text attribute for indented blocks [GH-794][]
+    - Some sources of pathological behavior of markdown-match-bold and markdown-match-italic
+      on large blocks have been mitigated
 
 *   Bug fixes:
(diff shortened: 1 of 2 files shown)


umanwizard:
This fixes at least some of the issues contributing to https://github.com/jrblevin/markdown-mode/issues/799 , although I think there is more slow behavior lurking, so I won't claim that issue is entirely fixed.

hmelman:
Thanks, this does fix a lot of #391 for me.

Issue #847, opened 2024-10-23
Wiki link markdown hiding not working


jayemar:
## Expected Behavior

If I have a wiki link such as `[[foo|bar]]` and I run `markdown-toggle-markup-hiding`, only `bar` should be shown.

This works for inline markdown links as well as reference links, but not for wiki links.  I have the value of `markdown-enable-wiki-links` set to `t`.

## Actual Behavior

Running the command `markdown-toggle-markup-hiding` has no effect on the display of the wiki link.  The entire link is shown both before and after enabling markup hiding:
`[[foo|bar]]`

## Steps to Reproduce

1. Enable wiki links: `(setq markdown-enable-wiki-links t)`
2. Run the command `markdown-insert-wiki-link`.  This will insert a pair of nested square brackets: `[[]]`
3. Type "foo|bar" inside of the square brackets so you now have `[[foo|bar]]`
4. Run the command `markdown-toggle-markup-hiding`.  The displayed text doesn't change.  Run the command again and it still doesn't change.

## Backtrace

There is no error

## Software Versions

markdown version 2.7-alpha
Emacs: 29.4
OS: Fedora 40
Examples:

## Additional Information

As I was looking through the code trying to better understand what's been happening I notice that wiki links seem to be handled much differently than inline markdown links and reference links, even for image links.  The other link types seem to be routed through much the same code, but wiki links have their own separate code path. Can you help me understand why this is?

Issue #782, opened 2023-07-28
Make wiki-links clickable by mouse like normal urls/links


xot:
## Expected Behavior

Make wiki-links like [[file]] (pointing to a markdown file in the current directory) behave exactly like normal links like https://www.example.com. I.e. make the font change when hovering over, and making them clickable by the mouse (opening the file like C-c C-o would)

## Actual Behavior

wiki links are for highlighted (when markdown-enable-wiki-links is t), but are not 'active' as normal links are.

## Steps to Reproduce

Create the following markdown file:

[[file]]

http://www.example.com

## Software Versions

- Markdown Mode: 2.5
- Emacs: 28.2
- OS: MacOS 13.4.1

Issue #557, opened 2020-11-01
Clickable wiki links


saf-dmitry:
For the sake of usability and consistency I would suggest making wiki links clickable same as normal Markdown links already are.

## Software Versions

- Markdown Mode: From GitHub on 2020-11-01
- Emacs: 26.3
- OS: macOS 10.14.6

PR #861, opened 2024-11-17
Hide wiki markup as part of markdown-toggle-markup-hiding

jayemar:
When markup hiding is enabled, the markup for markdown-style links is hidden, but the markup for wiki links is not.  This code change fontifies the different sections of wiki links so that they are properly supported by `markdown-toggle-markup-hiding`.

## Related Issues

The changes address issues #557, #782, and #847.

## Type of Change

## Checklist

- [X] I have added an entry to **CHANGES.md**.

### DIFF
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -10067,6 +10067,20 @@ rows and columns and the column alignment."
                    (propertize link 'face 'markdown-reference-face)
                    (propertize "]" 'face 'markdown-markup-face))
                 (propertize link 'face 'markdown-url-face)))))
+   ;; Hidden URL for wiki links
+   ((and (and markdown-enable-wiki-links
+              (thing-at-point-looking-at markdown-regex-wiki-link))
+         (or markdown-hide-urls markdown-hide-markup))
+    (let* ((imagep (string-equal (match-string 1) "!"))
+           (referencep (string-equal (match-string 5) "["))
+           (part1 (match-string-no-properties 3))
+           (part2 (match-string-no-properties 5))
+           (link (if markdown-wiki-link-alias-first part2 part1))
+           (edit-keys (markdown--substitute-command-keys
+                       "\\[markdown-insert-wiki-link]"))
+           (edit-str (propertize edit-keys 'face 'font-lock-constant-face)))
+      (format "Hidden URL (%s to edit): %s"
+              edit-str (propertize link 'face 'markdown-reference-face))))
    ;; Hidden language name for fenced code blocks
    ((and (markdown-code-block-at-point-p)
          (not (get-text-property (point) 'markdown-pre))
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -262,7 +262,7 @@ This is the default search behavior of Ikiwiki."
   :package-version '(markdown-mode . "2.2"))
 
 (defcustom markdown-wiki-link-search-type nil
-  "Searching type for markdown wiki link.
+  "A list of options for searching for markdown wiki links.
 
 `sub-directories': search for wiki link targets in sub directories
 `parent-directories': search for wiki link targets in parent directories
@@ -8386,6 +8386,11 @@ The location of the alias component depends on the value of
     (or (match-string-no-properties 5) (match-string-no-properties 3))))
 
 (defun markdown--wiki-link-search-types ()
+  "Return a list of the currently selected search types.
+
+Due to deprecated variables, as of markdown-mode version 2.5, the return value
+of this function is the same as the value of the variable
+`markdown-wiki-link-search-type'."
   (let ((ret (and markdown-wiki-link-search-type
                   (cl-copy-list markdown-wiki-link-search-type))))
     (when (and markdown-wiki-link-search-subdirectories
@@ -8397,6 +8402,7 @@ The location of the alias component depends on the value of
     ret))
 
 (defun markdown--project-root ()
+  "Try various approaches to find the project root."
   (or (cl-loop for dir in '(".git" ".hg" ".svn")
                when (locate-dominating-file default-directory dir)
                return it)
--- a/markdown-mode.el
+++ b/markdown-mode.el
@@ -1980,12 +1980,14 @@ inline code fragments and code blocks."
 
 (defface markdown-link-face
   '((t (:inherit link)))
-  "Face for links."
+  "Face for link text, ie the alias portion of a link.
+If a wiki link does not have an alias portion, use this face
+for the entire link."
   :group 'markdown-faces)
 
 (defface markdown-missing-link-face
   '((t (:inherit font-lock-warning-face)))
-  "Face for missing links."
+  "Face for the link text if the link points to a missing file."
   :group 'markdown-faces)
 
 (defface markdown-reference-face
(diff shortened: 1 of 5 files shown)


markdown-mode.el
```diff
@@ -8174,6 +8229,33 @@ Translate filenames using `markdown-filename-translate-function'."
           (markdown--browse-url (markdown-link-url)))
     (user-error "Point is not at a Markdown link or URL")))
 
+(defconst markup-props (list 'invisible 'markdown-markup
```
syohex, line 8232: Please add package prefix to global variables according to Emacs lisp naming convention

syohex:
- Please fix all byte-compile warnings
- Please add unit tests

markdown-mode.el
jayemar: I fixed this and ran a linter to help catch some other issues

jayemar:
I fixed the byte compiler warnings and updated the wiki link tests to check that the face properties are properly being set to support markup hiding. I think I've done everything that's called out in `CONTRIBUTING.md`, but if not let me know.

Wiki links seemed to be handled very different front inline and reference links.  I couldn't find a reason for the two different approaches, and it make some of the implementation more difficult than it needed to be, so some of my changes have the effect of treating wiki links much more like reference and inline links.  If I misunderstood this and there's a reason for the two different approaches please let me know.  (That's only if `markdown-enable-wiki-links` is `t`; otherwise they'll still ignored.)

jayemar:
I see this workflow is halted with the message that it is awaiting approval from a maintainer.  Is there anything I can do or need to do in order to get this branch through the build pipeline?

jayemar:
@syohex fixed merge conflicts and failing tests from recent updates to mouse face in links

markdown-mode.el
```diff
@@ -3394,23 +3395,6 @@ the buffer)."
 (defun markdown-match-yaml-metadata-key (last)
   (markdown-match-propertized-text 'markdown-metadata-key last))
 
-(defun markdown-match-wiki-link (last)
```
syohex, line 3397: Please don't remove public interfaces, this causes breaking backward compatibility

markdown-mode.el
```diff
@@ -8451,39 +8510,6 @@ See `markdown-wiki-link-p' and `markdown-follow-wiki-link'."
       (markdown-follow-wiki-link (markdown-wiki-link-link) arg)
     (user-error "Point is not at a Wiki Link")))
 
-(defun markdown-highlight-wiki-link (from to face)
```
syohex, line 8454: Please don't remove public interfaces, this causes breaking backward compatibility

markdown-mode.el
```diff
@@ -8451,39 +8510,6 @@ See `markdown-wiki-link-p' and `markdown-follow-wiki-link'."
...
-  "Highlight the wiki link in the region between FROM and TO using FACE."
-  (put-text-property from to 'font-lock-face face))
-
-(defun markdown-unfontify-region-wiki-links (from to)
```
syohex, line 8458: Please don't remove public interfaces, this causes breaking backward compatibility

markdown-mode.el
```diff
@@ -8451,39 +8510,6 @@ See `markdown-wiki-link-p' and `markdown-follow-wiki-link'."
...
-  "Remove wiki link faces from the region specified by FROM and TO."
-  (interactive "*r")
-  (let ((modified (buffer-modified-p)))
-    (remove-text-properties from to '(font-lock-face markdown-link-face))
-    (remove-text-properties from to '(font-lock-face markdown-missing-link-face))
-    ;; remove-text-properties marks the buffer modified in emacs 24.3,
-    ;; undo that if it wasn't originally marked modified
-    (set-buffer-modified-p modified)))
-
-(defun markdown-fontify-region-wiki-links (from to)
```
syohex, line 8468: Please don't remove public interfaces, this causes breaking backward compatibility

markdown-mode.el
```diff
@@ -8502,50 +8528,6 @@ newline after."
         (setq new-to (point)))
     (cl-values new-from new-to)))
 
-(defun markdown-check-change-for-wiki-link (from to)
```
syohex, line 8505: Please don't remove public interfaces, this causes breaking backward compatibility

markdown-mode.el
```diff
@@ -8502,50 +8528,6 @@ newline after."
...
-                  ;; Unfontify existing fontification (start from scratch)
-                  (markdown-unfontify-region-wiki-links new-from new-to)
-                  ;; Now do the fontification.
-                  (markdown-fontify-region-wiki-links new-from new-to))))))
-      (cursor-intangible-mode -1)
-      (and (not modified)
-           (buffer-modified-p)
-           (set-buffer-modified-p nil)))))
-
-(defun markdown-check-change-for-wiki-link-after-change (from to _)
```
syohex, line 8539: Please don't remove public interfaces, this causes breaking backward compatibility

markdown-mode.el
```diff
@@ -8502,50 +8528,6 @@ newline after."
...
-  "Check region between FROM and TO for wiki links and re-fontify as needed.
-Designed to be used with the `after-change-functions' hook."
-  (markdown-check-change-for-wiki-link from to))
-
-(defun markdown-fontify-buffer-wiki-links ()
```
syohex, line 8544: Please don't remove public interfaces, this causes breaking backward compatibility

markdown-mode.el
```diff
@@ -8556,33 +8538,10 @@ and disable it otherwise."
             (not markdown-enable-wiki-links)
           (> (prefix-numeric-value arg) 0)))
   (when (called-interactively-p 'interactive)
-    (message "markdown-mode wiki link support %s" (if markdown-enable-wiki-links "enabled" "disabled")))
+    (message "markdown-mode wiki link support %s"
+             (if markdown-enable-wiki-links "enabled" "disabled")))
   (markdown-reload-extensions))
 
-(defun markdown-setup-wiki-link-hooks ()
```
syohex, line 8562: Please don't remove public interfaces, this causes breaking backward compatibility

syohex (commented):
Please don't remove public interfaces  all of sudden.  When we remove public interfaces, fist we have to set them as obsoleted function/variable. Then after several version releases, then we will be able to remove them

jayemar:
@syohex Thank you for the feedback.  You're right that I should not have removed public functions so I've added them back in.

However, these functions were originally called via a hook that is no longer necessary as the markup for wiki links is now handled via `markdown-mode-font-lock-keywords` so as to function like other link types.  Given this, should I mark these functions as obsolete?

syohex:
> should I mark these functions as obsolete?

If it is no longer necessary or we can use another one instead of them, then please mark them as obsoleted variables/functions.

markdown-mode.el
```diff
@@ -278,6 +278,16 @@ project: search for wiki link targets under project root"
 (make-obsolete-variable 'markdown-wiki-link-search-subdirectories 'markdown-wiki-link-search-type "2.5")
 (make-obsolete-variable 'markdown-wiki-link-search-parent-directories 'markdown-wiki-link-search-type "2.5")
```
jayemar, line 280: These functions are no longer used due to the way wiki links are handled with the changes

jayemar:
@syohex I have marked the functions that are no longer used/necessary as obsolete

markdown-mode.el
```diff
@@ -278,6 +278,16 @@ This is the default search behavior of Ikiwiki."
...
 
+(make-obsolete 'markdown-match-wiki-link nil "2.3")
```
syohex, line 281: The 3rd argument of the `make-obsolete` means that **when** this function was made obsolete. So, it should be next release version(`2.8`), not past released versions

syohex:
Please update CHANGES.md

jayemar:
I've updated CHANGES.md to reference the changes from this branch.

And thanks for the changes to fix the byte compiler warnings!

PR #876, opened 2025-01-17
Add defun markdown-is-fontify-buffer-p

shipmints:
This is extremely useful in a `prog-mode' hook to avoid resource-intensive features such as `eglot' inside a fontification buffer.

## Type of Change

## Checklist


shipmints:
I'm rescinding this PR in favor of recommending the following which can be used in prog-modes to test for avoiding expensive resources when in non-file buffers or conventionally hidden buffers:
```elisp
    (when (or buffer-file-name
              (not (string-prefix-p " " (buffer-name))))
      ;; do something expensive
      )
```
Perhaps we can add this to the documentation. I'll make a PR for that.

PR #877, opened 2025-01-18
Document markdown-toggle-fontify-code-blocks-natively usage hint

shipmints:
This can help users avoid expensive resource invocations in their prog-mode hooks for fontification buffers.


shipmints:
A mitigation is rather than

```elisp
(funcall mode)
```
instead do this
```elisp
(delay-mode-hooks (funcall mode))
```
But that runs the risk of missing font-lock and other user preferences in their mode hooks.  It's a tradeoff.

Issue #822, opened 2024-01-23
Bug: Angle-bracketed URLs in inline code are fontified as links


saf-dmitry:
## Expected Behavior

Bare URLs and URL-like expressions in angled brackets (e.g., `<https://google.com>` or `<svg:tspan>`) _inside inline code spans_ are fontified as inline code.

## Actual Behavior

Bare URLs and URL-like expressions in angled brackets (e.g., `<https://google.com>` or `<svg:tspan>`) _inside inline code spans_ are fontified as links:

![Screen Shot 2024-03-25 at 10 42 07](https://github.com/jrblevin/markdown-mode/assets/30050534/d5c4b31b-3a3e-4efa-a94e-2d511e20f408)

## Steps to Reproduce

1. Input the following text:

    ```
    `<https://google.com>`
    `<svg:tspan>`
    ```
2. See `<https://google.com>` and `<svg:tspan>` fontified as links.

## Backtrace

n.a.

## Software Versions

- Markdown Mode: From GitHub on 2023-10-18
- Emacs: 28.2.1
- OS: macOS Catalina

## Additional info

Possibly broken after the 0a6a8d85934f8a8d289c43316a07b179498d9ac1 commit.

saf-dmitry:
Is fontification of angle-bracketed URLs _everywhere_ probably a desired behavior?

Hi-Angel:
Thanks! I created a PR that should fix this issue https://github.com/jrblevin/markdown-mode/pull/878

PR #878, opened 2025-02-11
Don't break inline-code rendering when there's a <foo:bar>

Hi-Angel:
## Description

A Markdown like `value=<addr:port>` results in `<addr:port>` suddenly refusing to highlight. This is because the mode recognizes this as an angle URL. Generally speaking, a URL (even an actual one) is not expected to be highlighted inside inline code-blocks, so this commit adds a test for whether start or end is in inline block, to avoid breaking rendering.

A test for this is added as well.

## Related Issue

Fixes: https://github.com/jrblevin/markdown-mode/issues/822

## Type of Change

## Checklist


syohex:
Thanks

PR #899, opened 2025-04-29
Add * to the markdown-mode syntax table as punctuation

jamescherti:
This commit adds `*` to the markdown-mode syntax table as punctuation. The asterisk is widely used in markdown for emphasis (bold and italic).

## Description

This commit adds `*` to the markdown-mode syntax table as punctuation. The asterisk is widely used in markdown for emphasis (bold and italic).

## Related Issue

Here are an example of issues that this pull request fixes:
- `dabbrev` includes `*` when completing a word surrounded by asterisks.
- The built-in `highlight-symbol-at-point` function highlighting the symbol at the point includes `*`.

## Type of Change

## Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@
     - Support drag and drop features on Windows and multiple files' drag and drop
     - Added cmark and cmark-gfm to the markdown command list.
     - Disable `imenu-submenus-on-top` by default [GH-882][]
+    - Add $%*+/<=>_|&' to the markdown-mode syntax table as punctuation.
 
   [gh-847]: https://github.com/jrblevin/markdown-mode/issues/847
   [gh-861]: https://github.com/jrblevin/markdown-mode/pull/861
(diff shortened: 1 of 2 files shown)


jamescherti:
Originally, I wanted to add more than `*` this pull request. For example:
```
>
<
-
_
```

Let me know if you think it would be a good idea to add more of them to this pull request, such as the `rst.el` package. They could improve editing in markdown modes.

I believe we can confidently use `rst.el` as a reference to guide what should be included in this pull request.

syohex:
I think that markdown-mode should set other characters to correct syntax properties. So please add configurations to characters other than `*`.

BTW I'm wondering why org-mode doesn't set syntax properties of markup characters.

jamescherti:
> I think that markdown-mode should set other characters to correct syntax properties. So please add configurations to characters other than `*`.

Sounds good, @syohex. I'll add more characters to this pull request alongside `*` and keep you updated.

> BTW I'm wondering why org-mode doesn't set syntax properties of markup characters.

I was wondering the same thing myself...

PR #948, opened 2026-08-20
Fix `gfm-mode` exception when doc ends with unterminated code fence

rhgg2:
## Description

When the buffer's final line is an opening code fence with no trailing newline, the `markdown-gfm-block-begin` property extends to `point-max`, so `next-single-property-change` returns nil and `goto-char` signals.

If the property reaches the end of the buffer, there can be no further blocks to visit, so setting point to `point-max` is safe: on the next iteration, `markdown-find-next-prop` returns nil, and the loop terminates without further ado.

## Type of Change

## Checklist


syohex:
Thanks for fixing the issue

